PR #6, opened 2014-09-14
Add a draft for indexes.

mjtamlyn:
Some draft API specifications for custom indexes. Key points:

- Introduction of `django.db.models.indexes`
- Support for custom classes to `db_index`
- Introduction of `Meta.indexes`
- `spatial_index` would be deprecated and a corresponding chunk of code removed from gis, allowing fields to specify their own index type, via `Field.get_default_index`

### DIFF
--- a/drafts/indexes.rst
+++ b/drafts/indexes.rst
@@ -1,0 +1,116 @@
+Advanced indexes
+================
+
+:Created: 2014-09-14
+:Author: Marc Tamlyn
+:Status: Draft
+
+Abstract
+========
+
+This DEP aims to rework how Django handles index definitions. This will allow
+Django to support more complex indexes on the relevant backends (most notably
+PostgreSQL), such as functional indexes and indexes with different algorithms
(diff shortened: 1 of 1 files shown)


jacobian:
At this point this is still missing a lot of the basic formatting and team required by the DEP process, so I'm going to close this PR. If and when a team emerges, please feel free to re-open or open a new PR with this DEP. Thanks!

auvipy:
possibilities for class based index's in next django release? probably 1.10?

jdufresne:
In PR django/django#5766 for [trac ticket #20581](https://code.djangoproject.com/ticket/20581), @charettes mentioned that this DEP could be a possible solution for introducing a mechanism allowing for deferrable unique constraints. Are there plans to address this as part of this proposal? I see this as an additional motivation for richer index/constraint syntax.

mjtamlyn:
I don't see any reason why this couldn't be supported by a more powerful indexes/constraints API.

PR #12, opened 2015-03-07
DEP 191, 192: Composite fields

ovangle:
Provided first drafts of the proposals relative to composite fields
See https://groups.google.com/forum/?utm_source=footer#!topic/django-developers/MZUcOE6-7GY
for the mailing list discussion.

Some of the syntax of these DEPs are incorrect. Should be suitable
for a first draft though. I'll update as I go along.

### DIFF
--- a/draft/191-composite-fields.rst
+++ b/draft/191-composite-fields.rst
@@ -8,8 +8,8 @@ DEP 191: Composite Fields
 :Shepherd: __
 :Status: Draft
 :Type: Feature
-:Created: 2015-03-06
-:Last-Modified: 2015-03-06
+:Created: 2015-03-12
+:Last-Modified: 2015-03-12
 
 .. contents:: Table of Contents
    :depth: 3
@@ -21,8 +21,8 @@ Abstract
 Define a new type of virtual field which maps onto one or more concrete
 database fields in the model's table. These fields can be used to define
 table constraints which span multiple columns with greater locality than if the
-same constraints were created using the current `index_togther` or
-`unique_together` APIs.
+same constraints were created using the current ``index_togther`` or
+``unique_together`` APIs.
 
 Specification
 =============
@@ -30,47 +30,55 @@ Specification
 CompositeField
 --------------
 
-A `CompositeField` is a new virtual field type which combines multiple columns
+A ``CompositeField`` is a new virtual field type which combines multiple columns
 on a model's table into a single object, which is accessible from model
 instances.
 
-The value of a `CompositeField` on a model instance is either, depending on
+The value of a ``CompositeField`` on a model instance is either, depending on
 
-* A dict, mapping subfield values to their values on the model
-* A tuple of the field's subfield values
+* A ``dict``, mapping subfield values to their values on the model (the
+  default); or
+* A ``tuple`` of the field's subfield values, ordered with the default subfield
+  ordering.
 
-depending on the value of `field.is_tuple_value`.
+depending on the value of ``field.value_is_tuple``.
+
+Limiting the field to either dicts or tuples seems restrictive, but supporting
+arbitrary python objects is proposed as part of DEP 192.
 
 Subfield
 --------
 
-A subfield is a proxy for another field. Subfields are ordered by the time
-they were added to the composite field.
-
-The value for a subfield on a model instance is the same as the value
+A ``subfield`` of a composite field is a delegate for another field on the same
+model. The value for a subfield on a model instance is the same as the value
 for the proxied field.
 
+Subfields are ordered firstly by the execution order of calling the ``__init__``
+method of the composite field, and then within the composite field by the
+execution order of the proxied field.
+
+
+Descriptor
+----------
+
+Since a composite field does not directly store a value on the model instance,
+a python descriptor which delegates getting/setting the instance value of the
+subfields is available from
 
 
 Providing a constraint that spans multiple columns
 --------------------------------------------------
-
-*FIXME*:
-    Does it make sense to create table indexes using `constrain`? Can we
-    support `CHECK` constraints too?
 
 A new top-level function will be added to the models API with the following
 signature
 
 .. code:: Python
 
-    def constrain(*fields, unique=False, index=True, is_tuple=False)
+    def constrain(*fields, unique=False, index=True, value_is_tuple=False)
 
-The `constrain` function will create a `CompositeField` in the model and add
+The ``constrain`` function will create a `CompositeField` in the model and add
 subfields to the composite field in the order in which they appear in the
 argument list.
-
-eg.
 
 .. code:: python
 
@@ -78,52 +86,106 @@ eg.
         x = models.IntegerField()
         y = models.IntegerField()
 
-        point = models.constrain(x, y, unique=True, is_tuple=True)
+        point = models.constrain(x, y, unique=True, value_is_tuple=True)
 
-This code inserts a composite field, with the name `point` to the model. A
+This code inserts a composite field, with the name ``point`` to the model. A
 constraint which ensures the uniqueness of the point will be added to the table
-and an index will be added to the
+and an index will be added to support improved lookups on the ``point`` object.
 
-.. note::
 
-    Composite fields provide a property which allows the value of the composite
-    field directly on the model, and a point can be provided to the model's
-    `__init__` method.
+*Note*: In this initial implementation, only ``UNIQUE`` constraints and ``INDEX``
+statements are supported. However the same technique should be available to
+create ``CHECK`` constraints in databases which support the functionality, as
+well as multi-column primary keys.
 
-    Calling the init method with both a value for the composite field _and_
-    a value for the subfield will raise a `ValueError`
 
-    e.g.
+Composite fields provide a property which allows the value of the composite
+field directly on the model, and a point can be provided to the model's
+``__init__`` method.
 
-    .. code:: python
+Calling the init method with both a value for the composite field _and_
+a value for the subfield will raise a ``ValueError``
 
-        >>> m = MyModel(x=4, point=(5,6))
+.. code:: python
+    >>> m = MyModel(x=4, point=(5,6))
 
-        ValueError: Multiple values for field 'x' (via keyword arguments 'x'
-                    and 'point')
+    ValueError: Multiple values for field 'x' (via keyword arguments 'x'
+                and 'point')
 
 Queries
 -------
 
-The values of a composite field must be queryable via the `Model.objects` API.
+The values of a composite field must be queryable via the ``Model.objects`` API.
 
 :`in`: Query for whether the value of the composite field is present in the
         provided list of values
-:`EXACT`: Query for whether the value of the composite field equals the provided
+:`exact`: Query for whether the value of the composite field equals the provided
           value
+:`isnull`: Query for whether the value for the composite field is `None`. It is
+           assumed
 
 In addition, for each of the subfields of the composite field, a transform will
 be provided which allows the user to perform a query on the subfield via the
 composite field.
 
-e.g.
-
-..code:: python
+.. code:: python
 
     MyModel.objects.filter(point__x__lt=4)
 
 would be transformed into a lookup of all point values which have an x value
 less than 4.
+
+
+Interactions with ``Model._meta``
+--------------------------------
+
+Since all subfields of a model are also local concrete fields of a model, they
+are available via ``Model._meta.fields`` with the appropriate subfield name.
+
+Since the composite field is virtual, it is not included in ``Model._meta.fields``,
+but can be accessed via ``Model._meta.virtual_fields``. It is also possible to
+retrieve a ``CompositeField`` instance via the method ``Model._meta.get_field``.
+
+To access the ``Subfield`` instance associated with the field (rather than the
+wrapped field itself), the composite field exposes a ``get_subfield`` method.
+
+For example, given the following model:
+
+..code:: python
+
+  >>> class MyModel(models.Model):
+  ...    x = models.IntegerField()
+  ...    y = models.IntegerField()
+  ...
+  ...    point = models.constrain(x, y, unique=True)
+  ...
+  >>> MyModel.get_field('x')
+  <IntegerField instance at 0x????????>
+  >>> MyModel.get_field('point')
+  <CompositeField instance at 0x????????>
+  >>> _.get_subfield('x')
+  <Subfield instance at 0x????????>
+
+An additional argument ``include_subfields`` which defaults to ``False`` will
+be added to the ``Options.get_fields`` method, which will include all subfield
+instances of the model.
+
+
+..code:: python
+
+  >>> for f in MyModel._meta.get_fields():
+  ...     print(f)
+  <IntegerField MyModel.x>
+  <IntegerField MyModel.y>
+  <CompositeField MyModel.point>
+  >>> for f in MyModel._meta.get_fields(include_subfields=True):
+  ...     print(f)
+  <IntegerField MyModel.x>
+  <IntegerField MyModel.y>
+  <Subfield MyModel.point.x>
+  <Subfield MyModel.point.y>
+  <CompositeField MyModel.point>
+
 
 
 Motivation
--- a/draft/191-composite-fields.rst
+++ b/draft/191-composite-fields.rst
@@ -1,0 +1,157 @@
+=========================
+DEP 191: Composite Fields
+=========================
+
+:DEP: 191
+:Author: Thomas Stephenson
+:Implementation Team: Thomas Stephenson
+:Shepherd: __
+:Status: Draft
+:Type: Feature
+:Created: 2015-03-06
+:Last-Modified: 2015-03-06
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+Abstract
+========
+
+Define a new type of virtual field which maps onto one or more concrete
+database fields in the model's table. These fields can be used to define
+table constraints which span multiple columns with greater locality than if the
+same constraints were created using the current `index_togther` or
+`unique_together` APIs.
+
+Specification
+=============
+
+CompositeField
+--------------
+
+A `CompositeField` is a new virtual field type which combines multiple columns
+on a model's table into a single object, which is accessible from model
+instances.
+
+The value of a `CompositeField` on a model instance is either, depending on
+
+* A dict, mapping subfield values to their values on the model
+* A tuple of the field's subfield values
+
+depending on the value of `field.is_tuple_value`.
+
+Subfield
+--------
+
+A subfield is a proxy for another field. Subfields are ordered by the time
+they were added to the composite field.
+
+The value for a subfield on a model instance is the same as the value
+for the proxied field.
+
+
+
+Providing a constraint that spans multiple columns
+--------------------------------------------------
+
+*FIXME*:
+    Does it make sense to create table indexes using `constrain`? Can we
+    support `CHECK` constraints too?
+
+A new top-level function will be added to the models API with the following
+signature
+
+.. code:: Python
+
+    def constrain(*fields, unique=False, index=True, is_tuple=False)
+
+The `constrain` function will create a `CompositeField` in the model and add
+subfields to the composite field in the order in which they appear in the
+argument list.
+
+eg.
+
+.. code:: python
+
+    class MyModel(models.Model):
+        x = models.IntegerField()
+        y = models.IntegerField()
+
+        point = models.constrain(x, y, unique=True, is_tuple=True)
+
+This code inserts a composite field, with the name `point` to the model. A
+constraint which ensures the uniqueness of the point will be added to the table
+and an index will be added to the
+
+.. note::
+
+    Composite fields provide a property which allows the value of the composite
+    field directly on the model, and a point can be provided to the model's
+    `__init__` method.
+
+    Calling the init method with both a value for the composite field _and_
+    a value for the subfield will raise a `ValueError`
+
+    e.g.
+
+    .. code:: python
+
+        >>> m = MyModel(x=4, point=(5,6))
+
+        ValueError: Multiple values for field 'x' (via keyword arguments 'x'
+                    and 'point')
+
+Queries
+-------
+
+The values of a composite field must be queryable via the `Model.objects` API.
+
+:`in`: Query for whether the value of the composite field is present in the
+        provided list of values
+:`EXACT`: Query for whether the value of the composite field equals the provided
+          value
+
+In addition, for each of the subfields of the composite field, a transform will
+be provided which allows the user to perform a query on the subfield via the
+composite field.
+
+e.g.
+
+..code:: python
+
+    MyModel.objects.filter(point__x__lt=4)
+
+would be transformed into a lookup of all point values which have an x value
+less than 4.
+
+
+Motivation
+==========
+
+Django's model API provides a relatively coarse level of data abstraction,
+relying upon assumption that a single userland object will map to a single
+table in the database.
+
+In addition, this API provides the groundwork for adding data abstraction over
+column subsets (DEP 192) and the future implementation of multi-column primary
+keys.
+
+
+Backwards Incompatability
+=========================
+
+Deprecation of Model.Meta.index_together and Model.Meta.unique_together?
+
+Reference Implementation
+========================
+
+TBA
+
+Copyright
+=========
+
+This document has been placed in the public domain per the Creative Commons
+CC0 1.0 Universal license (http://creativecommons.org/publicdomain/zero/1.0/deed).
+
+(All DEPs must include this exact copyright statement.)
--- a/draft/191-composite-fields.rst
+++ b/draft/191-composite-fields.rst
@@ -5,7 +5,7 @@ DEP 191: Composite Fields
 :DEP: 191
 :Author: Thomas Stephenson
 :Implementation Team: Thomas Stephenson
-:Shepherd: __
+:Shepherd: Anssi Kääriäinen
 :Status: Draft
 :Type: Feature
 :Created: 2015-03-12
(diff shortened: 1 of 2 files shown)


draft/192-standalone-composite-fields.rst
```diff
@@ -0,0 +1,208 @@
...
+    TypeError: At most one class in the inheritance heirarchy of B can define a
+               subfield.
+
+Null Handling
+-------------
+
+Often, it is desirable to allow a `CompositeField` to be associated with a
+python `None` value, while enforcing `NOT NULL` constraints on the columns.
+If the `null` field parameter is passed to a `CompositeField` constructor, an
+implicit `field__isnull` column into the model's table and this value will be
```
ar45: I don't think there should be a third null param on the CompositeField. if any SubFields is not nullable than the CompositeField is not null. Setting a python None value to the CompositeField should be handled by the value_from_dict method.

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,157 @@
...
+--------------
+
+A `CompositeField` is a new virtual field type which combines multiple columns
+on a model's table into a single object, which is accessible from model
+instances.
+
+The value of a `CompositeField` on a model instance is either, depending on
+
+* A dict, mapping subfield values to their values on the model
+* A tuple of the field's subfield values
```
akaariai: I don't think it is a good idea to restrict to tuple or dict only. User defined objects which either support indexing or dict lookups should be supported, too.

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,157 @@
...
+
+depending on the value of `field.is_tuple_value`.
+
+Subfield
+--------
+
+A subfield is a proxy for another field. Subfields are ordered by the time
+they were added to the composite field.
```
akaariai: I'm not sure what this is about.

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,157 @@
...
+for the proxied field.
+
+
+
+Providing a constraint that spans multiple columns
+--------------------------------------------------
+
+*FIXME*:
+    Does it make sense to create table indexes using `constrain`? Can we
+    support `CHECK` constraints too?
```
akaariai: While Constraint as a class might make sense I think it will be easiest to just use regular CompositeFields for this case, too. We will need query support for unique constraints, and doing so requires a large part of the field API to be present.

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,157 @@
...
+                    and 'point')
+
+Queries
+-------
+
+The values of a composite field must be queryable via the `Model.objects` API.
+
+:`in`: Query for whether the value of the composite field is present in the
+        provided list of values
+:`EXACT`: Query for whether the value of the composite field equals the provided
```
akaariai: isnull is needed, too.

draft/192-standalone-composite-fields.rst
```diff
@@ -0,0 +1,208 @@
...
+----------------
+
+All of the parameters defined by `Field` and `CompositeField` are accepted
+by a standalone composite field with the following provisos:
+
+* `null` is an acceptable argument for a standalone composite field
+  (see :Handling python None:)
+* `is_tuple`, from composite field is an invalid argument. A standalone
+  field should define the value transformation functions `value_to_dict` and
+  `value_from_dict`, rather than auto-coercing the value into a tuple
```
akaariai: Again, I think it is better to leave a bit more room for other types, too. For example you might want to return Money instances that know how to convert themselves to other currencies.

The DEPs should contain minimal requirements for the return type.

akaariai:
Some specification of the structure of the compositefield would be useful. It seems that a composite field includes its subfields, *and a descriptor*. I'd also like to see how .values('point') should behave.

Field ordering should be specified. Maybe this:
  - Subfields must always be in the same class as the composite field, and must precede the composite field.
  - Autocreated subfields are placed immediately before the composite field in the field listing.

draft/191-composite-fields.rst
ovangle: Agreed, but that's what the overridable methods `value_from/to_dict` are for. I was looking for a simple API that covers the use case of an "inline" declaration of a composite field, and 

* dicts by default
* tuples if preferred

would cover that use case adequately. 

If you have a custom object you can define a custom field for that object using the standalone API in about 10 lines of code with the bonus that it is reusable in any model.

draft/191-composite-fields.rst
ovangle: Looks like something I started writing and got distracted by something else. I didn't really edit this stuff as well as I could have and I'm considering this an iterative process.

draft/191-composite-fields.rst
ovangle: `constrain` isn't a class, it's a factory function for `CompositeField`. So fields constructed with this method will still be regular `CompositeField`s.

draft/191-composite-fields.rst
ovangle: It's impossible for an inline field to be `None`. It will always be a tuple or a dict (with possibly null values, depending on the nullability of the constrained fields). 

DEP192 adds the requirement that `isnull` be a supported query, and since they share the same implementation inline fields will support it too, but that's more by accident than design.

draft/192-standalone-composite-fields.rst
ovangle: Yep -- standalone field will support user defined objects, with python functionality via the overridable methods `value_from/to_dict`.

ovangle:
I'll add some mention of the descriptor and expand upon subfields a bit more.

Subfield ordering is already specified:

> Subfields are ordered first by the execution time they are created

Although that could probably be changed to:
> Subfields are ordered first by the execution time that the composite field is created, and then by the execution time of the subfield.
which would be more accurate.

It would be a `NameError` to define the subfield after you refer to it in `constrain`. I don't think the specification should cover details that are prohibited by the language. 

Also, you can't put autocreated (managed?) subfields immediately before the composite field in the field listing. Model._meta adds virtual fields at the end of the field listing, possibly after any other non-composite virtual fields. Whether or not that's the intended behaviour is debatable, but it would be a backwards incompatible change to order virtual fields correctly :(

Subfields are inserted into the field listing _where the composite field would have been if it had been a concrete field_ could probably be added.

draft/192-standalone-composite-fields.rst
```diff
@@ -0,0 +1,218 @@
+=========================
+DEP 191: Composite Fields
```
slurms: 192? Same just below this.

ovangle: c+v error. Thanks and fixed

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,219 @@
+=========================
+DEP 191: Composite Fields
+=========================
+
+:DEP: 191
+:Author: Thomas Stephenson
+:Implementation Team: Thomas Stephenson
+:Shepherd: __
```
akaariai: Feel free to add me as Shepherd.

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,219 @@
...
+=============
+
+CompositeField
+--------------
+
+A ``CompositeField`` is a new virtual field type which combines multiple columns
+on a model's table into a single object, which is accessible from model
+instances.
+
+The value of a ``CompositeField`` on a model instance is either, depending on
```
akaariai: depending on duplicated

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Limiting the field to either dicts or tuples seems restrictive, but supporting
+arbitrary python objects is proposed as part of DEP 192.
+
+Subfield
+--------
+
+A ``subfield`` of a composite field is a delegate for another field on the same
+model. The value for a subfield on a model instance is the same as the value
+for the proxied field.
```
akaariai: For example, if you do this:
```
    obj.first_name = 'Anssi'
    obj.last_name = 'Kääriäinen'
    obj.full_name = ('Matti', 'Kääriäinen')
```
Then, what exactly happens when you set obj.full_name. If the value is a dict, and you do:
```
    obj.full_name['first_name'] = 'Matti'
```
then what should happen? Finally, we have the same situation for object based composite field values for DEP 192; what should happen when you do
```
    obj.full_name.first_name = 'Matti'
```

The rules should be made clear in this DEP, and it should be also mentioned how those rules are going to be implemented.

We have similar situation for ForeignKeys. It would be great if obj.book.id would always match obj.book_id, but that isn't the case currently. So, *not* requiring the same values always is also a possibility, but if this is chosen, then there should be clear reasoning why requiring same values always is impractical.

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,219 @@
...
+
+Subfields are ordered firstly by the execution order of calling the ``__init__``
+method of the composite field, and then within the composite field by the
+execution order of the proxied field.
```
akaariai: I don't get what this whole section means.

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,219 @@
...
+instances.
+
+The value of a ``CompositeField`` on a model instance is either, depending on
+
+* A ``dict``, mapping subfield values to their values on the model (the
+  default); or
+* A ``tuple`` of the field's subfield values, ordered with the default subfield
+  ordering.
+
+depending on the value of ``field.value_is_tuple``.
```
akaariai: value_is_tuple isn't perhaps the best API. Some other alternatives:
  - CompositeDictField, CompositeTupleField (and CompositeField base type for DEP-192)
  - value_type=tuple or value_type=dict

draft/191-composite-fields.rst
```diff
@@ -0,0 +1,219 @@
...
+  ...    point = models.constrain(x, y, unique=True)
+  ...
+  >>> MyModel.get_field('x')
+  <IntegerField instance at 0x????????>
+  >>> MyModel.get_field('point')
+  <CompositeField instance at 0x????????>
+  >>> _.get_subfield('x')
+  <Subfield instance at 0x????????>
+
+An additional argument ``include_subfields`` which defaults to ``False`` will
```
akaariai: I'm not sure we need this. Maybe it is better to leave this out of this DEP, and implement this separately if a use case arises. Of course, if you can provide an use case that isn't better solved by other means, then we can include this in this DEP directly.

draft/191-composite-fields.rst
ovangle: Will do.

draft/191-composite-fields.rst
ovangle: Setting the value of the subfield via the composite field value should overwrite any previous value for the subfields. It's an imperative language and we can't store multiple values for subfields, so...

The only time that this wouldn't be sensible behaviour during object initialization (since there is no sensible order on args/kwargs) and I've covered that elsewhere by saying that it is a `ValueError: Multiple values received for 'subfield_name'` if both a composite field and a subfield are specified in the model's `__init__` method.

Arbitrary python objects are a different beast though. In `191` we have the liberty of returning custom `tuple` and `dict` types which delegate `__getitem__` and `__setitem__` to mutations of the field values, but in `192` the field declaration is completely separate from the instance, so we can't override `__getattr__` and `__setattr__` accordingly.

Thus, `obj.full_name = Name('Matti', 'Kääriäinen')` would override the subfield values, but `obj.full_name.first_name = 'Matti' wouldn't. I hadn't really thought about that much, but it's obviously not ideal.

Possible solutions are:
- Mandating that objects with an associated field type extend an `ObservableMixin` (with 
  implementation details TBA because it would need some experimentation). 

  This would keep the field and object definitions separate, allowing users to declare field types for objects defined in third party libraries (although they'd still have to subclass them to add the `ObservableMixin` base). 
- Merge declarations of objects and their associated fields, which would be simpler, but restricts field types to classes defined by the user and adds a lot of irrelevant methods to the API of the objects. 

- ?? (Open to suggestions)

draft/191-composite-fields.rst
ovangle: A delegate is an object which takes its implementation from another object. The subfield doesn't actually have a value on the model, it's value is the same as whichever field it is delegating to.

Re: ordering, all concrete field types are currently ordered temporally by the time of execution that their `__init__` method is called (the `creation_counter`). Composite fields are "ordered" temporally with other concrete fields (even though they are virtual and virtual field aren't ordered) and subfields share their `creation_counter` with the composite field and are then ordered by the creation order of the delegate field.

So, for example:

```class MyModel(models.Model):
    x = models.SomeField()
    y = models.SomeField()
   point = models.constrain(x, y)
   z = models.SomeField()
```

Statements are executed in order inside a class definition, so:
```x < y < point < z```
Subfields share the creation counter of point and `x < y`, so they are ordered in the field list between `point` and `z`. 

```x < y < point < point__x < point__y < z```

This is a sensible definition, but can cause some problems with tuples:

```
class MyModel(...):
   y = models.SomeField()
   x = models.SomeField()
   point = models.constrain(x, y)
```
The ordering of elements within tuples would be the same as the ordering of the composite field, so 
```
>>> my_model.x = 4
>>> my_model.y = 5
>>> print(my_model.point)
(5,4)
```

Which is counter-intuitive, but if you start overwriting the creation counters of subfields based on `inspect.getargs` or something, well... The temporal ordering by execution time is consistent with the way fields in django have been traditionally ordered. It's why I made the default implementation of `constrain` return a `dict` (and only added `value_is_tuple` after mailing list discussion).

File: draft/191-composite-fields.rst (hunk 12 above)
Comment (ovangle) [inline draft/191-composite-fields.rst]: See above note about ordering of subfields and not being enthusiastic about `tuple`s in `191` in the first place. 

But given the choice I'd choose the former and get rid of the `constrain` factory method.

File: draft/191-composite-fields.rst (hunk 13 above)
Comment (ovangle) [inline draft/191-composite-fields.rst]: As mentioned above, `Subfield`s are delegates for other fields. Including them in `_meta.get_fields` would cause "duplication" of information which might be unexpected if the user doesn't explicitly opt in to seeing them.

As for _why_ they are needed, the simplicity of implementation is the main justification. I don't like the idea of adding a whole bunch of `if self.subfield_creation_counter > 0` onto the `Field` base class, it's much easier to create a delegate and leave `90%` of the logic related to implementing them in the `composite.py` module.

Generally, the user can just interact with the fields directly and ignore the fact that the fields are actually boxed by the composite field implentation. It will be a very rare case for the user to need to call `get_fields` with `include_subfields=True`.

**Comment by ovangle** [comment]
@akaariai 
Thanks for the comprehensive review, you've given me a lot to think about. I'll wait for your responses before updating the DEPs appropriately.

**Comment by akaariai** [comment]
I don't see the need for delegate subfields. This seems like a non-warranted complication to me.

In addition, exactly what happens with the duplicate values on obj.first_name = 'Anssi'; obj.full_name.first_name = 'Anssi' should be considered extremely carefully. Discussing this on the django-developers mailing list could be a good idea.

**Comment by ovangle** [comment]
I'm just going to get on with my life. I posted to django-dev _with an 80% complete reference implementation_. You can take that and modify it any way you like, but I'm not going to be told that I don't know what I'm talking about when it comes to "non-warranted complications". 

And "discuss it on the mailing list"? What the hell? I was told to open up a DEP because it would be a better forum for discussing a change like this. I gave two perfectly reasonable suggestions as to how to deal with the problem.

**Comment by akaariai** [comment]
I wasn't telling you that you don't know what you are doing. But I don't see any evidence that delegate fields are actually needed. I can't spot why the delegate fields are needed from the reference implementation.

A reference implementation isn't any proof by itself, there are multiple ways to implement this feature, and we want to make sure we have the best implementation. Michal Petrucha implemented composite fields *without* delegate fields, and quickly checking his implementation doesn't have too many complications. So I stand by my point: by the evidence I have delegate fields seem to be non-needed complication.

Telling to discuss this on the mailing list wasn't a good idea. The main point is that the community should have a consensus on the way setting values on composite fields and its subfields works. The tradeoffs should be clearly documented in this DEP. I think this is the most important point when it comes to this DEP. What we choose we have to live with, and it is easy to make a mistake here.

**Comment by ovangle** [comment]
You're right, and I perhaps acted a bit harshly to being told to go back to the mailing list for discussion. I shouldn't have closed the PR, but I've made an effort to clearly justify every point I've made and your response was clearly lacking in detail, in essence being:

> I don't like the idea of subfields 

why? what are the pros and cons of removing them from the proposal? Why are they an unwarranted "complication" and what makes them a "complication" at all?

> Attribute access should be carefully considered

Yes I agree, so let's consider them. But saying we should consider them is just bureaucratic cruft and doesn't move the conversation forwards in any way, shape or form.



There are indeed multiple ways of implementing anything and yes, all the logic related to subfields can be put on the already bloated `Field` base type without complicating matters too much. But there is a clear gain as far as I'm concerned to codifying the difference between a field which is attached to a composite field and a field which is attached to the model.

Pros of having a subfield class:
- It makes reasoning about when a field should be contributed to a model and whether the model or the composite field is responsible for management of the field easier. 
- It also avoids mutations of the creation counter
- It allows subfields to be named in a way which is prohibited by the base field type. 
- Subfield specific checks don't need to be added to the `Field` type.

Cons of having a subfield class
- It adds an extra class which has only superfluous differences to the field base class.

**Comment by akaariai** [comment]
I'll take a look if there is some other solution to the delegate fields. In any case, keeping them as undocumented implementation detail seems to be a good idea.

For the attribute access I think the following is a credible alternative.

Subfields on the model and subfields on the composite field's value have descriptors. Whenever they are assigned a value, the matching value on the "other side" is updated, too. So, for example, if you assign to foo.first_name, then the descriptor also sets foo.full_name.first_name. When you assign to foo.full_name.first_name, then foo.first_name is changed, too.

The above will be technically feasible to do. There is some coding to get this implemented. There will also be a performance penalty as setting attributes through descriptor is a bit slower than without descriptors. On the other hand having the same value live in two different places in an unsynchronized way *will* lead to a lot of bugs and confusion.

We have also the option that we only implement class based composite fields, and the composite field doesn't store any value at all itself (at least not when the composite field is assigned to a model), it is just a proxy for the value on the model. So, foo.full_name.first_name will go through full_name's first_name descriptor that is implemented as: 
```
if self.instance:
    return getattr(self.instance, self.field.name)
else:
    return getattr(self, self.field_name)
```
similarly for the setter.

This way it will be *impossible* to have the values to be out of sync. Unfortunately this also means we can't have tuple or dict based composite fields, or that we have different rules for those.

I'm not sure which way is the best. There are certainly some tradeoffs with each approach.

**Comment by ovangle** [comment]
Regarding subfields: this isn't supposed to be documentation, it's supposed to be specification, so I'll leave it in there for the moment. I'll remove the ability to fetch subfields from `Meta.get_fields()`, so they are completely opaque to the user for the moment. It can be added again later if deemed necessary. 

Regarding custom objects, I'll review the current options, with hope to making a correct decision here, because I feel it's important.

This is what I had originally planned for defining a field based on a custom object:

    class Money(object):
       def __init__(self, code, amount):
          self.code = code
          self.amount = amount

    class MoneyField(CompositeField):
       code = CharField(...)
       amount = DecimalField(...)

       def value_to_dict(self, value):
          return dict(getattr(value, attr) for attr in {'code', 'amount'})

      def value_from_dict(self, value):
          return Money(**value)

Then when defined on a field

    class StockItem(Model):
        money = MoneyField()

Now, when you're using a custom object, the `money__code` and `money__amount` fields which are contributed to the `StockItem` model are not documented implementation details. There should be no reason to worry about "synchronising" their values, since if a user is accessing them, they are doing so in an officially unsupported way.

But there is a problem with this approach:

    stock_item = StockItem(money=Money('USD', Decimal('4.02'))
    stock_item.money.code = 'AUD'

The attribute access `stock_item.money` returns a completely new instance of the `Money` object and sends it off into user land. However we have no way of observing this new instance once it has been created, so the assignment to `code` has no effect on the model. 

Here are the options that have been proposed to deal with this circumstance:

## 1. ObservableMixin

The first choice is that we have 

    class ObservableMixin(object):
       def _set_context(self, field, model):
          self._model = model
          self._field = field

       def __getattribute__(self, attr):
           if field.name == attr:
               return self._field.get_field_value(model)
           try:
               return self.__dict__[attr]
           except KeyError:
               raise AttributeError(...)
 
       def __setattr__(self, attr):
           if self._field.name == attr:
               return self.field.get_field_value(model)
          try:
              self.__dict__[attr] = value
          except KeyError:
              raise AttributeError(...)

Then the `Money` object would be defined with `class Money(ObservableMixin)`

pros

- Completely transparent, works with no knowledge of what is going on behind the scenes
- Allows _any_ type of custom object to be used as a composite field type.

cons

- Slightly magic
- Assumes that attributes on the custom object have the same name as the subfields (could be fixed by adding a `source` on the subfield definition).

## 2. Merge objects and their composite fields

Rather than defining a separate object and field, the field and the object share the same definitions

    class Money(CompositeField):
        code = CharField(...)
        amount = DecimalField(...)

Then we have complete control over how attributes are accessed off the model and can get and can reflect any change to an instance value back onto the field without worrying about stale data. 

pros:

- Simple format
- Familiar, since entirely reflects model syntax
- no `value_to_dict` or `value_from_dict` methods necessary. 

cons:

- Limits composite field types to those defined by the user
- Mixes the static and dynamic definitions of the instance (one of the fundamental problem with models in django)
- A choice must be made over whether it is more important for instances to have customisable behaviour or `CompositeField`s to have customisable behaviour of `CompositeField` instances (instances win).
- A custom field instance can't be created without being attached to a model (so you can't have the `Money` instances in the wild without an associated `StockItem`).
         
## 3. Merge objects and field definitions and add descriptors on the instance field

Basically equivalent to `2.`, except instead of directly setting the value of the field on the instance, it is set via a descriptor.

pros:
All the pros of `2.`
cons:
All the cons of `2.` plus a performance penalty.   


I think that covers all the options. `dict`s that are returned by `191` will be a custom `dict` implementation which override `__getitem__` and `__setitem__` to refer to the values of the delegate fields, so they are still supportable in any of the above cases.

**Comment by ovangle** [comment]
I should note that `ObservableMixin` is by far my preferred option at the moment, although I'm prepared to be convinced otherwise.

**Comment by akaariai** [comment]
I prefer ObservableMixin, too. I think the implementation should be descriptor based, not `__getattr__` or `__getattribute__` based. But lets not document the exact implementation details in this DEP, this leaves room to test different implementation approaches later on.

I think the implementation must have the following characteristicts:
  1. When the composite field isn't assigned with a model, the composite field itself stores assigned values.
  2. When the composite field is assigned with a model, the composite field  proxies the values of the model.
  3. When a composite field is assigned to a different model, the values "travel" with the composite field.

So, in code, 1 means that:
```
fullname = FullName()
fullname.firstname = 'Anssi'
fullname.lastname = 'Smith'
obj.fullname = fullname  # at this point firstname and lastname are copied to obj
obj.fullname.firstname = 'Matti'  # assigns to obj.firstname and caches the value in obj.firstname
```
2 means that:
```
obj.fullname.firstname  # goes through descriptor or getattr on fullname to fetch value from obj directly
obj.fullname.firstname = 'Jack'  # sets value on obj
```
3 means that:
```
obj.fullname = FullName('Anssi', 'Smith')
obj.fullname.firstname = 'Jack'
obj.lastname = 'Wilson'
obj2.fullname = obj.fullname.copy()
obj.firstname == 'Jack'
obj.lastname == 'Wilson'
```
There are two reasons for the copy operation:
  1. If we don't have that, then obj.lastname assignment would need to cache the value on obj.fullname
  2. We have to copy the fullname instance in any case, as the same instance can't proxy two different models.

I don't see a fundamental reason why all composite field implementations must obey these rules. For example 3rd party projects might do something else entirely. But this seems like a good, safe and reasonably fast default implementation.

akaariai:
We have at least one more issue that needs to be dealt with in this DEP - how do we serialize and deserialize composite values? Typical use case is what to use as a key for composite field in HTML <select> elements.

Koniiiik's GSoC work has a solution for this issue, I think we can just copy that.

akaariai:
Oh, wait. I'm not at all sure I like the ObservableMixin idea. Because from the code presented I don't see how that actually works. I think I need to see a bit more (how the composite field is defined in different use cases, and how the model definition using that composite field works like).

Also, reading some of the argumentation above, we can't rely on the concrete fields on the main model to be implementation defaults. For example if you define a unique constraint as a composite field, you likely will still use the concrete fields directly.

ovangle:
> Because from the code presented I don't see how that actually works

The code presented is just wrong :D (I wrote it to sketch out the idea, not as an example). It is possible to refactor it to use descriptors, however there is a distinct danger to using descriptors and that is the possibility of shadowing the descriptors on the implementation class.

suppose we have

ovangle:
ps. I haven't actually opened an editor and tested the `__getattribute__` stuff. I know `__getattribute__` on a class will be called before looking at the namespace of the class, but I'm not entirely sure what happens when it is defined on a subclass. My expectation is *exactly the same thing will happen, otherwise `__getattribute__` would  be entirely useless*.

There's also a possibility of breaking if a subclass overrides `__getattribute__` but you could build an infinite improbability drive with the chances of that causing a problem.

akaariai:
Descriptors wouldn't work all that well in any case. What I'm thinking is that when a composite field is assigned to a model, then changes to the composite field's values do not change anything on the composite field itself. So, descriptors wouldn't have a chance to do anything at all.

Also, we could add a check that we are not overriding any descriptor.

Serialization of composite fields is done somewhere here: https://github.com/django/django/compare/master...koniiiik:soc2013/composite-fields. (Although I'm not entirely sure which branch contains the latest code). You *really* should check what was done by koniiiik. The code was almost ready for merge, so it contains a *lot* of useful information about the problematic cases for composite fields.

Lets see if we can get something useful done without serialization to strings. The smaller the required changes are the better.

ovangle:
Oh, the GSOC for composite fields. Yes, I am already aware of it but the reference to 'koniiiks GSOC work' passed over my head. 

I'll update the proposal with decisions when I get a chance over the next couple of days. Then hopefully we can accept and move it to drafts, so people can raise specific issues using the deps issue tracker rather than adding more comments to the PR.

jacobian:
@ovangle this looks like it's got all the pieces to be merged as a draft DEP, except a [shepherd](https://github.com/django/deps/blob/master/final/0001-dep-process.rst#forming-the-team). So the next step is to find a core dev who is interested in shepherding the proposal. I'll see if I can help there, but if and when you get to that point feel free to ping me and I'll merge this to drafts.

ovangle:
@jacobian,

@akaariai nominated himself as shepherd (somewhere) in the discussion above, I added his name to the deps and have been waiting for a reply since I last updated the proposals, however none has been forthcoming. 

I'm quite fine in updating them again with a different dev if he has rescinded his offer or if he is ineligible/occupied, but he's been quite proactive in discussing the proposal so far and I've valued his input.

jacobian:
@ovangle I'm sorry, I must have somehow missed that -- that's my bad. OK, let's get this out of a PR!

PR #13, opened 2015-04-13
Update JavaScript Tests DEP

treyhunner:
I am reviving this DEP draft and I plan to work on implementation during the PyCon sprints over the next few days.

### DIFF
--- a/draft/javascript-tests.rst
+++ b/draft/javascript-tests.rst
@@ -24,13 +24,10 @@ Create unit tests for the JavaScript code in the admin and gis contrib packages.
 Specification
 =============
 
-A native JavaScript test framework is one that can be run without any Python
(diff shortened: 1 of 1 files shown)


carljm:
Thanks! This DEP meets the basic formatting requirements, and is technically coherent, which are the only requirements for merge into Draft status. So merging this PR and assigning a DEP number.

In addition to working on implementation, it would be good to mail django-developers and let people know you are working on this.

I'll also mention that while I am willing to shepherd this DEP (I do have experience with JS testing tools), I'm happy to hand off that role to @jezdez or @jphalip or anyone else who has done more actual work in the Django JS, if they want the role.

treyhunner:
Great!  I will notify the django-developers mailing list soon.

PR #19, opened 2016-03-03
Add draft DEP: Advanced Indexes

akki:
A proposal for class based indexes.
Based on @mjtamlyn's originial proposal on Advanced Indexes.

### DIFF
--- a/draft/0019-Advanced-Indexes.rst
+++ b/draft/0019-Advanced-Indexes.rst
@@ -1,0 +1,124 @@
+==========================
+DEP 0019: Advanced Indexes
+==========================
+
+:DEP: 0019
+:Author: Marc Tamlyn, Akshesh Doshi
+:Status: Draft
+:Type: Feature
+:Created: 2016-03-03
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+Abstract
+========
+This proposal aims to bring flexibility in Django on what it can do with db
+indexes and allow Django to support more complex indexes on relevant backends
+(most notably PostgreSQL), such as functional indexes and indexes with different
+algorithms (e.g. ``GiST``).
+
+Motivation
+==========
+
+There are new fields being introduced in ``contrib.postgres`` which have
+similar issues to the rationale for ``spatial_index``, a ``BTree`` index is
+either insufficent, inefficient or does not work and an alternative index type
+is needed for that field. A field should be able to declare the type of index which
+is appropriate to it.
+
+Also indexes need to be made more customisable, like to be able to extend
+support for `DEFERRABLE INITIALLY DEFERRED for UNIQUE constraints
+<https://code.djangoproject.com/ticket/20581>`_.
+
+In addition, both postgres and oracle support powerful functional indexes
+(and mysql supports indexing on the first few characters of a string), which
+allow significant performance benefits for certain queries. Of course, it is
+possible to create these indexes by hand if you know what you are doing using
+``RunSQL`` based migrations but this is a place for improvement.
+
+PostgreSQL also supports partial indexes.
+
+With the recent advances in expression syntax, they can be extended to indexes
+with modifications in the way indexes work.
+
+Proposed solution and specifications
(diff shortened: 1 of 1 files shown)


draft/0019-Advanced-Indexes.rst
```diff
@@ -0,0 +1,124 @@
...
+with modifications in the way indexes work.
+
+Proposed solution and specifications
+====================================
+
+Introduce ``django.db.models.indexes`` with deconstructable ``Index`` class.
+This will be inherited by different index algorithms such as ``BTree`` and ``Hash``,
+and also ``FunctionalIndex(expression)``.
+
+``Index`` classes will take 3 optional arguments ``fields``, ``model`` and ``name``.
```
MarkusH: This paragraph is unclear to me. maybe give the exact function signature you want to implement?

draft/0019-Advanced-Indexes.rst
```diff
@@ -0,0 +1,124 @@
...
+that for spatialite (using a ``SELECT`` statement and a function) without
+having to subclass ``SchemaEditor``.
+
+Introduce ``Meta.indexes`` for other indexes. This will contain a list of
+``Index()`` instances. ``index_together`` will continue to be supported but
+will be translated into ``Meta.indexes`` internally and handled only there by
+``SchemaEditor``.
+
+The general workflow for index addition/deletion would be -
+migrations framework detects that Meta.indexes has changed => models layer
```
MarkusH: The models layer is not involved here. The autodetector constructs the migration operations which at the end of `makemigrations` are then written out into the migration files.

draft/0019-Advanced-Indexes.rst
```diff
@@ -0,0 +1,124 @@
...
+    class Town(models.Model):
+        name = models.CharField(max_length=255)
+        slug = models.SlugField(unique=True)
+        region = gis.PolygonField(db_index=True)
+        area = models.FloatField()
+        population_density = models.FloatField()
+
+        class Meta:
+            indexes = [
+                models.BTree(fields=['slug', self.name]),
```
MarkusH: I think that's not what we agreed on in our last GSOC call. My notes have

```python
class Index(object):
    def __init__(self, *fields, **options):
        pass
```

draft/0019-Advanced-Indexes.rst
```diff
@@ -0,0 +1,124 @@
...
+                models.FunctionalIndex(expression=(F('area') * F('population_density')),
+                    type=models.BTree),
+            ]
+
+Backwards compatibility
+=======================
+
+``spatial_index`` will be deprecated as setting ``db_index`` on a spatial field
+will be sufficient.
```
MarkusH: We need to think of a way where the current implementation of `db_index=True` yields a `BTree` but a future default index class of a field might be `GiST`. When a user runs `manage.py makemigrations` they will get a `CreateIndex` operation that will probably try to drop a `GiST` index because that's what the default for that field is.

draft/0019-Advanced-Indexes.rst
akki: @MarkusH - This is a very old version of the proposal.
You can find the current implementation [here](https://github.com/django/django/pull/6726/files#diff-803276386305dab54c20b97d3d34b82fR12).
Thanks.

draft/0019-Advanced-Indexes.rst
akki: auto-detector will detect that `Meta.indexes` has changed (using something like `generate_altered_indexes`) and will update the `model_state` accordingly.

akki:
Closing due to inactivity and lack of updation from my side as well. Other discussions can be found [here](https://groups.google.com/forum/#!topic/django-developers/lASYirbQvQA).

draft/0019-Advanced-Indexes.rst
MarkusH: That might be, but the DEP should be updated accordingly, IMO.

PR #20, opened 2016-05-08
[WIP] Channels DEP

jacobian:
TODO:

- [x] complete "motivation" section
- [x] complete "rationale" section
- [x] spel chek and proofreed

### DIFF
--- a/draft/0006-channels.rst
+++ b/draft/0006-channels.rst
@@ -27,32 +27,32 @@ particularly WebSockets — don't work in this simple round-trip way. Adding
 support for these protocols to today's Django apps is can be complicated and
 difficult.
 
-Thus, Channels. The top-line feature of Channels is simple, easy-to-use support
+Thus, channels. The top-line feature of channels is simple, easy-to-use support
 for WebSockets. Channels makes it very easy to write apps that support
-WebSockets alongside traditional HTTP views. Notably, Channels does not
+WebSockets alongside traditional HTTP views. Notably, channels does not
 introduce asyncio, gevent, or any other async code to Django app; all the code
 users write runs synchronously in a worker process or thread. This is primarily
 an accessiblity play: it's traditionally been very easy to write a "Hello World"
-HTTP view; the goal of Channels is to make it just that easy to write "Hello
+HTTP view; the goal of channels is to make it just that easy to write "Hello
 World" WebSockets views.
 
-Under the hood, Channels does a fair bit more. In a nutshell, Channels replaces
+Under the hood, channels does a fair bit more. In a nutshell, channels replaces
 Django's request/response cycle with *messages* that are sent across *channels*.
-This means that adding Channels to Django enabled not just WebSocket support,
+This means that adding channels to Django enabled not just WebSocket support,
 but support for any future protocol (whether uni- or bi-directional). Channels
 also allows for background tasks to run out-of-band of the web servers, similar
 to some of the feaures of Celery or Python-RQ (though its featureset is far
 smaller, and it is not intended to replace most uses of task queues)
 
 Although this is a fairly profound change to Django's internals, it's also fully
-backwards-compatible. Django will continue to run under WSGI, and Channels
+backwards-compatible. Django will continue to run under WSGI, and channels
 barely touches the WSGI codepaths. Thus, the new features only "appear" when
-running Django in "Channels mode".
+running Django in "channels mode".
 
 Specification
 =============
 
-The section summarizes the basic concepts of Channels, shows a simple example,
+The section summarizes the basic concepts of channels, shows a simple example,
 and explains the implementation plan. This is intended as a high-level overview;
 for detailed specifications, documentation, and examples see `the Channels
 documentation <https://channels.readthedocs.io/>`_.
@@ -62,15 +62,19 @@ Channels Concepts
 
 The core of the system is, unsurprisingly, a datastructure called a *channel*.
 What is a channel? It is an *ordered, first-in first-out queue* with message
-expiry and at-most-once delivery to only one listener at a time.
+expiry, capacity, and at-most-once delivery to only one listener at a time.
 
 You can think of it as analogous to a task queue - messages are put onto the
 channel by producers, and then given to just one of the consumers listening to
 that channel.
 
-For more details, see `What is a Channel? <https://channels.readthedocs.io/en/latest/concepts.html#what-is-a-channel>`_.
+For more details, see `What is a channel? <https://channels.readthedocs.io/en/latest/concepts.html#what-is-a-channel>`_.
 
-To deploy an app using Channels, instead of running a WSGI server, you'll run a
+The other key concept is Groups, a set of channels that you can send messages
+to and get them delivered to all channels in the group. Groups are
+network-transparent and look the same across all 
+
+To deploy an app using channels, instead of running a WSGI server, you'll run a
 few different things:
 
 * An *interface server* that handles HTTP and WebSocket connections. 
@@ -108,10 +112,10 @@ few different things:
        implement workers in other Python web frameworks, such as Flask,
        and it may even be possible to implement workers in other languages.
 
-An example app using Channels
+An example app using channels
 -----------------------------
 
-To make this concrete, let's look at a simple example app using Channels.  The
+To make this concrete, let's look at a simple example app using channels.  The
 app is a simple real-time chat app — like a very, very light-weight Slack. There are
 a bunch of rooms, and everyone in the same room can chat, in real-time, with
 each other using WebSockets. This section will show off the highlights;
@@ -213,7 +217,7 @@ Integration plan
 We propose the following integration plan:
 
 * Merge `Channels <https://github.com/andrewgodwin/channels>`_ into Django 1.10.
-  Document the Channels APIs as "provisional" (using the terminalogy from
+  Document the channels APIs as "provisional" (using the terminalogy from
   `PEP 411 <https://www.python.org/dev/peps/pep-0411/>`_) so that we have room
   to make API changes. We think changes will be fairly unlikely -- the current
   design represents over two years of design work -- but we should leave the 
@@ -237,7 +241,7 @@ We propose the following integration plan:
 Motivation
 ==========
 
-The primary motivation for Channels is that of a percieved gap in Django's
+The primary motivation for channels is that of a percieved gap in Django's
 abilities; as the Web grows and evolves, the original view-based design has
 lasted surprisingly well, but is starting to chafe when presented with some
 of the new technologies the web is growing, particularly WebSockets.
@@ -253,14 +257,14 @@ their own, Django loses out on the potential for a community of apps, examples
 and code around WebSockets that has brought it as far as it has today for
 normal HTTP and view code.
 
-Thus, Channels' goal is to create a single, unified interface for Django
+Thus, channels' goal is to create a single, unified interface for Django
 developers to write their applications against (the consumer and routing model
 shown above), and to provide a good abstraction that allows extension and
 adaptation of the underlying coordination logic by end-users, specialists, or
 the project itself in the future (ASGI).
 
 Like the rest of Django, we cannot hope to satisfy everyone's needs, and in
-particular it is unlikely Channels could be used as-is at huge scale; however,
+particular it is unlikely channels could be used as-is at huge scale; however,
 no generic component survives that trip, and any resulting code always ends up
 very company- and situation-specific.
 
@@ -271,7 +275,7 @@ communication requirements, but the growth of existing integrations with other
 platforms like Slack provides ample opportunity for Django to position itself
 as an easy-to-use and reliable solution for all sorts of backend needs.
 
-The core Channels design is protocol-agnostic; while it ships with HTTP and
+The core channels design is protocol-agnostic; while it ships with HTTP and
 WebSocket support, work is either planned or already underway
 for Slack, IRC, email, HTTP/2 and SMS interface servers, allowing developers
 to use the same, familiar consumers-and-routing structure to service all kinds
@@ -296,7 +300,7 @@ as close as possible.
 Rationale
 =========
 
-There are several obvious alternatives to Channels that could be taken, and
+There are several obvious alternatives to channels that could be taken, and
 some major decisions in its design that have at first glance equally viable
 alternatives. This section tries to address some of the more important ones.
 
@@ -342,7 +346,7 @@ Channels' core abstraction, the channel, has at-most-once delivery. This choice
 is one side of a binary choice that all queue systems must make; at-most-once,
 or at-least-once.
 
-The situations that Channels will actually drop messages in are quite small;
+The situations that channels will actually drop messages in are quite small;
 mostly, they revolve around servers unexpectedly dying, or inordinate amounts
 of traffic filling up the channel capacity. In general, day-to-day use, users
 would likely see less than 0.01% of messages dropped.
@@ -366,7 +370,7 @@ and interface servers in the same deployment see the same channels and groups.
 
 This introduces what may seem like unnecessary complexity, but it addresses
 a key scaling problem that any project that grows past a single node must
-consider - broadcast. Many applications for Channels, such as chat systems,
+consider - broadcast. Many applications for channels, such as chat systems,
 notifications, live blogs and status GUIs, require the ability to send messages
 to an end-user WebSocket (or other open socket) from any number of places in
 the system - model code, consumers on other sockets, CLI tools, etc.
@@ -378,16 +382,16 @@ broadcast messages to large groups of connected sockets would likely have been
 very inefficient in terms of network traffic without the interface servers also
 understanding the network routing system at a higher level.
 
-Thus, the network transparency is built-in to Channels at the core, allowing
+Thus, the network transparency is built-in to channels at the core, allowing
 not only broadcast but a host of other useful features, like the ability to
 dedicate and tune machines to a single role (interface, worker, or worker on
 specific channels), and the lack of requirement for session stickiness.
 
 Small-scale deployments that only run on a single machine can still use a
-machine- or process-local channel backend, and Channels comes with one of each;
+machine- or process-local channel backend, and channels comes with one of each;
 scaling down is important, too.
 
-The ASGI specification, which defines the channel and group transport Channels
+The ASGI specification, which defines the channel and group transport channels
 uses, is designed to only impose as many guarantees and provide just enough API
 that it can be sensibly built against while allowing flexibility in
 implementation; writing a network-transparent channel layer is difficult, but
@@ -413,7 +417,7 @@ before others have finished executing.
 
 Solving this problem in a general way in a networked system is impossible to
 do without a significant performance hit, either by coordination or session
-stickiness. For this reason, Channels leaves the non-global-ordering,
+stickiness. For this reason, channels leaves the non-global-ordering,
 simultaneous style as the default, and provides a decorator, ``enforce_ordering``,
 that provides one of two levels of ordering and exclusivity guarantees at
 different levels of performance degredation.
@@ -439,7 +443,7 @@ Backwards Compatibility
 Channels is fully backwards-compatible. Until you switch into ASGI mode by
 deploying an interface server and running workers, Django continues to use
 the WSGI codepaths. This means that performance under WSGI is unchanged
-by the introduction of Channels.
+by the introduction of channels.
 
 The underlying architecture *does* change substantially after switching into
 ASGI mode, but that's an explicit opt-in step, and thus has no backwards-
--- a/draft/0006-channels.rst
+++ b/draft/0006-channels.rst
@@ -1,0 +1,284 @@
+==================
+DEP 0006: Channels
+==================
+
+:DEP: 0006
+:Author: Jacob Kaplan-Moss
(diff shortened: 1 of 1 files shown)


draft/0006-channels.rst
```diff
@@ -0,0 +1,468 @@
...
+   :local:
+
+Abstract
+========
+
+Historically, Django's request handing has been very simplistic, built around
+the concept of requests and responses: the browser makes a request, Django calls
+a view, and the view returns a response back to the browser.
+
+However, this architecture is starting to show its age. New protocols —
```
MarkusH: replace `—` with `--`.

draft/0006-channels.rst
```diff
@@ -0,0 +1,468 @@
...
+particularly WebSockets — don't work in this simple round-trip way. Adding
```
MarkusH: dito

draft/0006-channels.rst
```diff
@@ -0,0 +1,468 @@
...
+support for these protocols to today's Django apps is can be complicated and
```
MarkusH: chop "is"

draft/0006-channels.rst
```diff
@@ -0,0 +1,468 @@
...
+difficult.
+
+Thus, Channels. The top-line feature of Channels is simple, easy-to-use support
+for WebSockets. Channels makes it very easy to write apps that support
+WebSockets alongside traditional HTTP views. Notably, Channels does not
+introduce asyncio, gevent, or any other async code to Django app; all the code
```
MarkusH: Either `s/Django app/Django` or `s/to Django app/to a Django app`. I'd go with the first.

draft/0006-channels.rst
```diff
@@ -0,0 +1,472 @@
...
+
+However, this architecture is starting to show its age. New protocols —
+particularly WebSockets — don't work in this simple round-trip way. Adding
+support for these protocols to today's Django apps is can be complicated and
+difficult.
+
+Thus, channels. The top-line feature of channels is simple, easy-to-use support
+for WebSockets. Channels makes it very easy to write apps that support
+WebSockets alongside traditional HTTP views. Notably, channels does not
+introduce asyncio, gevent, or any other async code to Django app; all the code
```
MarkusH: Either `s/Django app/Django` or `s/to Django app/to a Django` app. I'd go with the first.

andrewgodwin:
As per the style of the docs in the channels PR, I have lowercased all mentions of "channels" referring to the Django module. References to the third-party app remain uppercased, as per its docs style.

draft/0006-channels.rst
```diff
@@ -0,0 +1,472 @@
...
+HTTP view; the goal of channels is to make it just that easy to write "Hello
+World" WebSockets views.
+
+Under the hood, channels does a fair bit more. In a nutshell, channels replaces
+Django's request/response cycle with *messages* that are sent across *channels*.
+This means that adding channels to Django enabled not just WebSocket support,
+but support for any future protocol (whether uni- or bi-directional). Channels
+also allows for background tasks to run out-of-band of the web servers, similar
+to some of the feaures of Celery or Python-RQ (though its featureset is far
+smaller, and it is not intended to replace most uses of task queues)
```
MarkusH: Add trailing period.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+
+.. [#] Like the interface server, it's possible for there to be different, 
+       non-Django worker implementations. It ought to be fairly trivial to
+       implement workers in other Python web frameworks, such as Flask,
+       and it may even be possible to implement workers in other languages.
+
+An example app using channels
+-----------------------------
+
+To make this concrete, let's look at a simple example app using channels.  The
```
MarkusH: Replace double spaces with single one.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+app is a simple real-time chat app — like a very, very light-weight Slack. There are
```
MarkusH: Replace `—` with `--`

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+a bunch of rooms, and everyone in the same room can chat, in real-time, with
+each other using WebSockets. This section will show off the highlights;
+for full details see `the code on Github <https://github.com/jacobian/channels-example>`_
+and `this article about the app on Heroku's blog <https://blog.heroku.com/archives/2016/3/17/in_deep_with_django_channels_the_future_of_real_time_apps_in_django>`_ [#]_
```
MarkusH: Line break before URL

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+
+.. [#] In the intersts of full disclusre, I should state that I (Jacob Kaplan-Moss)
```
MarkusH: "disclosure"

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+for full details see `the code on Github <https://github.com/jacobian/channels-example>`_
+and `this article about the app on Heroku's blog <https://blog.heroku.com/archives/2016/3/17/in_deep_with_django_channels_the_future_of_real_time_apps_in_django>`_ [#]_
+
+.. [#] In the intersts of full disclusre, I should state that I (Jacob Kaplan-Moss) 
+       was paid by Heroku to write the referenced blog post and example app.
+
+For the most part, this app looks like a normal Django app -- models, views,
+templates, URLs, etc. Let's look at the parts that are different:
+
+First, the app needs to define a channel layer in `settings.py`::
```
MarkusH: Use double back-ticks around `settings.py`

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+        "default": {
+            "BACKEND": "asgi_redis.RedisChannelLayer",
+            "CONFIG": {
+                "hosts": [os.environ.get('REDIS_URL', 'redis://localhost:6379')],
+            },
+            "ROUTING": "chat.routing.channel_routing",
+        },
+    }
+
+For more details on channel layers, see the `Channel Layer Types <https://channels.readthedocs.io/en/latest/backends.html>`_ docs.
```
MarkusH: Line break before URL

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+    from channels.routing import route
+    from . import consumers
+
+    channel_routing = [
+        route("websocket.connect", consumers.ws_connect),
+        route("websocket.receive", consumers.ws_receive),
+        route("websocket.disconnect", consumers.ws_disconnect),
+    ]
+
+For more details on channel routing, see the `Channel Routing <https://channels.readthedocs.io/en/latest/getting-started.html#routing>`_ docs.
```
MarkusH: Another line break, please.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+
+For more details on channel layers, see the `Channel Layer Types <https://channels.readthedocs.io/en/latest/backends.html>`_ docs.
+
+The channel layer points to our *channel routing* -- a structure that maps
+channel names to the functions that handle them::
+
+    # chat/routing.py
+
+    from channels.routing import route
+    from . import consumers
```
MarkusH: Can we add a blank line above, please.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+    ]
+
+For more details on channel routing, see the `Channel Routing <https://channels.readthedocs.io/en/latest/getting-started.html#routing>`_ docs.
+
+Here's what one of the consumers looks like::
+
+    # chat/consumers.py
+
+    import json
+    from channels import Group
```
MarkusH: Can we use Django's `isort` settings here, too, please.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+Notice that this looks fairly similar to an HTTP view, except that instead 
+of a request in recieves a message, and it doesn't return a response. Channels
+are uni-directional, so to send data back to the browser we need to send it
+on a *response channel*. In this case, we broadcast to a `group <https://channels.readthedocs.io/en/latest/getting-started.html#groups>`_, which takes care of sending to each
+user connected to the room. 
+
+For a full breakdown of these example consumers, see the 
+`websocket consumers section of the blog post <https://blog.heroku.com/archives/2016/3/17/in_deep_with_django_channels_the_future_of_real_time_apps_in_django#websocket-consumers>`_.
+
+Finally, we need to deploy this thing using ASGI instead of WSGI. To do that,
```
MarkusH: s/this thing/the chat app/

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+  design represents over two years of design work -- but we should leave the 
+  possibilty open.
+
+  This is implemented as `PR #6419 <https://github.com/django/django/pull/6419>`_.
+
+* Keep the other components -- `Daphne <https://github.com/andrewgodwin/daphne>`_,
+  `asgiref <https://github.com/andrewgodwin/asgiref>`_
+  and `asgi_redis <https://github.com/andrewgodwin/asgi_redis>`_ -- as
+  external components [#]_. Since these run independently of Django, they can be
+  iterated on separately from Django's release cycle.
```
MarkusH: Add an point to move `asgiref`, `asgi_redis`, `daphne` under the `django` github org?

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+something familiar, safe, and that performs reasonably well.
+
+At-most-once delivery
+---------------------
+
+Channels' core abstraction, the channel, has at-most-once delivery. This choice
+is one side of a binary choice that all queue systems must make; at-most-once,
+or at-least-once.
+
+The situations that channels will actually drop messages in are quite small;
```
MarkusH: "The situations in which channels will actually drop messages are quite small"

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+implementation; writing a network-transparent channel layer is difficult, but
+not tying Django to a single one and decoupling it like this allows both
+iteration on the one or two preferred solutions, and lets large companies or
+projects built out their own to suit their specific needs.
+
+Multi-listener ordering
+-----------------------
+
+While channels guarantees ordering of messages on a channel when there is a
+single listener - for example, when an interface server is reading a response
```
MarkusH: Use double `--` instead of `-`

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+body to send back to a connected client - it does not guarantee global ordering
```
MarkusH: dito

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+simultaneous style as the default, and provides a decorator, ``enforce_ordering``,
+that provides one of two levels of ordering and exclusivity guarantees at
+different levels of performance degredation.
+
+Alternatives
+------------
+
+There are many alternative architectures to the ones proposed by this DEP, and
+each has their advantages and disadvantages. Channels does not intend to make
+it impossible to use these; indeed, if someone wishes to run an
```
MarkusH: fix line wrap

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+
+* `asgi_redis <https://github.com/andrewgodwin/asgi_redis>`_ - Redis ASGI implementation
+
+Copyright
+=========
+
+This document has been placed in the public domain per the Creative Commons
+CC0 1.0 Universal license (http://creativecommons.org/publicdomain/zero/1.0/deed).
+
+(All DEPs must include this exact copyright statement.)
```
MarkusH: Drop this line?

MarkusH:
Thanks @jacobian nice work!

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+  The interface server is  roughly analgous to the role a WSGI server plays 
+  today. `Daphne <https://github.com/andrewgodwin/daphne/>`_ is the default
+  (production-ready) implementation [#]_, and for local development
+  ``python manage.py runserver`` can act as a (non-production-quality)
+  interface-server-and-worker-server-in-one.
+
+* Worker servers that run *consumers* that handle views, websockets, 
+  background tasks, etc. 
+
+  You'll run worker servers using ``python manage.py runworker`` [#].
```
MarkusH: Add trailing `_` to footnote reference.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+---------------------
+
+Channels' core abstraction, the channel, has at-most-once delivery. This choice
+is one side of a binary choice that all queue systems must make; at-most-once,
+or at-least-once.
+
+The situations that channels will actually drop messages in are quite small;
+mostly, they revolve around servers unexpectedly dying, or inordinate amounts
+of traffic filling up the channel capacity. In general, day-to-day use, users
+would likely see less than 0.01% of messages dropped.
```
mlavin: Most of this paragraph is a result of the implementation rather than the design itself. These statements have yet to be proven in the current implementation. I think it would be more accurate to say that limiting the number of lost messages and documenting potential failure cases is a goal for the resulting implementation.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+
+The choice of delivery guarantee informs the design of the rest of the solution,
+as well. With at-most-once, we will have to allow for retry logic and coding
+to cope with failure - something Django developers are very used to given the
+non-guaranteed nature of HTTP and browsers. If we were to have chosen
+at-least-once, however, we would have had to introduce a whole deduplication
+system and try and educate developers that their consumer code might be run
```
mlavin: The browsers/clients can already duplicate HTTP requests. Repeatedly clicking a form submit can produce multiple POSTs before the first response/redirect is returned to the browser. I can agree that this is a new wrinkle but the overall statelessness of HTTP requests makes it difficult to determine duplicates so I would say developers do have more experience than they are being given credit for here.

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+alternatives. This section tries to address some of the more important ones.
+
+In-process async/asyncio
+------------------------
+
+Python has had in-process async support for some time with solutions like
+Twisted and gevent, and with the introduction of ``asyncio`` in Python 3,
+an officially-blessed solution, too.
+
+Putting Django's Python 2 compatability requirement aside, the main argument
```
mlavin: Isn't Python 2 support scheduled to be removed in Django 2.0? If Python 2 support adds complication here then why not aim for that release?

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+WebSockets alongside traditional HTTP views. Notably, channels does not
+introduce asyncio, gevent, or any other async code to Django app; all the code
+users write runs synchronously in a worker process or thread. This is primarily
+an accessiblity play: it's traditionally been very easy to write a "Hello World"
+HTTP view; the goal of channels is to make it just that easy to write "Hello
+World" WebSockets views.
+
+Under the hood, channels does a fair bit more. In a nutshell, channels replaces
+Django's request/response cycle with *messages* that are sent across *channels*.
+This means that adding channels to Django enabled not just WebSocket support,
```
carljm: s/enabled/enables/

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+but support for any future protocol (whether uni- or bi-directional). Channels
+also allows for background tasks to run out-of-band of the web servers, similar
+to some of the feaures of Celery or Python-RQ (though its featureset is far
```
carljm: features

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+to and have them broadcast to all channels in the group. Groups are
+network-transparent and look the same across all machines running the same
+*channel layer*.
+
+To deploy an app using channels, instead of running a WSGI server, you'll run a
+few different things:
+
+* An *interface server* that handles HTTP and WebSocket connections. 
+
+  The interface server is  roughly analgous to the role a WSGI server plays
```
carljm: analogous

draft/0006-channels.rst
```diff
@@ -0,0 +1,473 @@
...
+before others have finished executing.
+
+Solving this problem in a general way in a networked system is impossible to
+do without a significant performance hit, either by coordination or session
+stickiness. For this reason, channels leaves the non-global-ordering,
+simultaneous style as the default, and provides a decorator, ``enforce_ordering``,
+that provides one of two levels of ordering and exclusivity guarantees at
+different levels of performance degredation.
+
+Alternatives
```
carljm: One specific alternative I'd like to see discussed here is the one @dstufft proposed, eliminating the message bus and network transparency in favor of just using `deferToThread` or equivalent to run sync "worker" code from an async server. Off the top of my head these are the pros/cons I'm aware of for that approach compared to channels (hopefully @dstufft or @andrewgodwin can correct these where I'm wrong or add ones I've missed):

- Pro: Simpler. No need to invent new solutions for large responses, streaming responses, message ordering. Less of a departure from how Django projects are served today.

- Pro: doesn't constrict communication between the interface server and workers, allowing richer feedback to flow in both directions, so server can make smarter decisions about load shedding and can kill long-running workers if the client connection goes away.

- Pro: less latency (though not clear yet how much less).

- Wash: Could be equally good as a low-barrier entry to writing websockets code with Django, without having to turn Django async from the ground up.

- Con: Doesn't provide any help with scaling out to multiple machines (especially scaling protocols with broadcast).

- Con: Requires everyone to use threads and write thread-safe code.

- Con (a huge one): Doesn't currently have a well-developed working implementation and a team/shepherd committed to working on it.

andrewgodwin: The major con is that it means Groups are impossible, and even letting people implement themselves by storing reply channel names is impossible; it prevents you being able to send to any channel from anywhere, you need to have some secondary mechanism to route things to the right machine to send them. It means we leave people out on their own to solve common patterns like notifying people about model saves or other events, and even makes implementing a chat impossible.

carljm:
Thanks @jacobian and @andrewgodwin. I'm doing a proof-read, spell-check, and formatting run-through and will push that shortly (please let me know if I'm duplicating un-pushed work!) After that I think this PR is ready for merge (which doesn't preclude future updates, of course); I'll wait for your confirmation before I merge it.

draft/0006-channels.rst
dstufft: > Doesn't provide any help with scaling out to multiple machines (especially scaling protocols with broadcast).

I'd suggest a rewording of this, the "simpler" idea just scales out like any WSGI application has scaled out (and similarly to how Channels scales out). Spin up more processes, the main difference is that with channels if you're already using redis from going 1 to N worker processes just requires spinning up the additional worker processes, but with the simpler thing it requires sticking something like HAProxy in front of it to load balance. However if you're using the in memory channels backend then it's basically a wash between needing to add HAProxy and needing to add Redis and if your front end (Daphne) needs to scale out then you're back at the HAProxy layer. IOW I think for scaling the simpler method is actually simpler in all but one specific scenario, you use a LB and add more processes, no matter how you're trying to scale.

The thing that I think this should probably address is that Channels provides built in support for actual broadcasting whereas the simpler thing will require bolting that on the side using Redis or some other pub sub thing.

carljm: Yep, agreed - I meant to capture that under the con of "scaling protocols with broadcast", but if this is incorporated into the DEP text your fuller explanation is much better.

dstufft: It makes implementing a chat ~~impossible~~ require additional tools.

draft/0006-channels.rst
carljm: @MarkusH I'm doing a proof-read run through, thanks for your eagle eyes! Just wanted to note that I'm not making this change; I think "this thing" here works fine as a colloquialism in a casual writing style.

draft/0006-channels.rst
carljm: @mlavin Sorry, this comment is now hidden in the diff due to my proof-read commit. (I'm not actually sure why, since I didn't change this particular line; I guess changing the context is enough?) Maybe you want to re-submit it?

draft/0006-channels.rst
carljm: Also this comment.

draft/0006-channels.rst
carljm: And this one.

carljm:
Pushed the proof-read commit. Things I see that still need attention:

1. Update the integration plan to adjust for the new plan to not merge into 1.10.

2. Responses to @mlavin 's comments (maybe integrate into the text in some form).

3. Add a summary of pro/cons relative to @dstufft alternative (I'm willing to do this based on Andrew and Donald's comments, unless someone else wants to).

I think all of these could happen as separate commits/PRs after merging this PR.

mlavin:
Regarding my comments, I think the Python 2 note is likely moot with the shift away from 1.10

MarkusH:
Thanks @carljm for incorporating my changes!

jacobian:
@carljm thanks for the proofs! I think we can go ahead and merge, the DEP doesn't need to be done-done to be on master, just complete-ish.

andrewgodwin:
I agree - I'd actually like to firm it up a bit now there's no need to rush it, and bring a bit more of the spec formally in, but it's definitely enough to go into master-drafts.

PR #23, opened 2016-06-08
Official projects DEP draft

andrewgodwin:
As discussed on django-developers, a DEP to formalise the process of adoption of project as "official Django projects" without being part of the core repo.

### DIFF
--- a/draft/0007-official-projects.rst
+++ b/draft/0007-official-projects.rst
@@ -38,14 +38,15 @@ increased exposure, existing project infrastructure and process (like the
 security reporting process), and lend them greater credibility for larger,
 more risk-averse companies and projects who aren't sure what option to pick.
 
-It also helps Django as a project overall; adopting external projects that
+It also helps Django as a project overall. Adopting external projects that
 solve problems and add features that the core Django project doesn't have
-helps us as a project by highlighting our feature-completeness as an overall
-community, aids new Django developers by highlighting projects that they might
-otherwise overlook or try and re-implement missing features themselves,
-and acts as a useful conduit to get people contributing to Django without
-having to tackle issues on the core project, which by this point are generally
-large, difficult issues that aren't as suitable for first-time contributors.
+helps us as a project:
+
+* It shows off Django-related features that people would otherwise miss.
+* It aids new Django developers by highlighting projects that they might
+  otherwise overlook.
+* It acts as a useful conduit to get people contributing to Django without
+  having to tackle (large, difficult) issues on the core project.
 
 Rationale
 =========
--- a/draft/0007-official-projects.rst
+++ b/draft/0007-official-projects.rst
@@ -223,6 +223,11 @@ to make it an official Django project:
   Shepherd and Maintenance Team, as well as details about Django's security
   policy.
 
+* The project's documentation should be updated to clarify how contributions
+  will be accepted (pull requests, patches, etc.) and what criteria the
+  Maintenance Team and Shepherd have for bringing on new contributors and
+  making them part of the Maintenance Team.
+
 * References to the project should be added in the official Django
   documentation where sensible, as well as other changes made to ensure it's
   discoverable.
@@ -246,6 +251,10 @@ project must maintain the following things:
 * Compatability with the current release and current LTS release of Django,
   within a month of the Django release coming out (LTS compatability may be
   with an older but still maintained version)
+
+* Responding to contributions made in the way suggested by the project's
+  contribution guide and updating the guide if the way they wish to accept
+  contributions changes.
 
 If any of these requirements does not continue to be true, effort should be
 made to find new maintainers or a new Shepherd to bring the project up to
--- a/draft/0007-official-projects.rst
+++ b/draft/0007-official-projects.rst
@@ -1,0 +1,282 @@
+==================================
+DEP 0007: Official Django Projects
+==================================
+
+:DEP: 0007
+:Author: Andrew Godwin
+:Implementation Team: Andrew Godwin
+:Shepherd: Andrew Godwin
+:Status: Draft
+:Type: Process
+:Created: 2016-06-01
+:Last-Modified: 2016-06-01
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+Abstract
+========
+
+Define a process for adopting and managing projects under the Django name
+and organisation that are not the core ``django`` repository and documentation.
+
+Motivation
+==========
+
+With the growing popularity of Django and the increasing number of projects
+living outside of the core repository that are nonetheless central to the
+Django community and mission, and a desire to not grow the size of the main
+Django repository with optional or extra code, it becomes apparent that there
+is a need for "official" Django projects that exist outside that repository
+but still are part of the Django organisation and management structure.
+
+Leaving these projects outside means they're often limited in their ability
+to maintain and grow the project based on the few people working on the
+project directly, and becoming an official project will benefit them with
+increased exposure, existing project infrastructure and process (like the
+security reporting process), and lend them greater credibility for larger,
+more risk-averse companies and projects who aren't sure what option to pick.
+
+It also helps Django as a project overall; adopting external projects that
+solve problems and add features that the core Django project doesn't have
+helps us as a project by highlighting our feature-completeness as an overall
+community, aids new Django developers by highlighting projects that they might
+otherwise overlook or try and re-implement missing features themselves,
+and acts as a useful conduit to get people contributing to Django without
+having to tackle issues on the core project, which by this point are generally
+large, difficult issues that aren't as suitable for first-time contributors.
+
+Rationale
+=========
+
+The goal of this proposed process is to enable a system where projects can
+be adopted and become official Django projects, but making sure it's not
+done prematurely (and thus stifling alternative approaches), and that it
+doesn't tie development to the release cadence and style of Django core
+releases too directly.
+
+That said, the Django name has gained a level of trust and respect over the
+years from our community and users, and so to become and official Django
+project a level of quality, commitment, and security must be met, and this
+process tries to ensure those are accounted for.
+
+Finally, projects and maintainers move on, and so this document tries to
+outline a process where projects can be deprecated or handed off between
+maintainers in a way that's transparent and allows Django as a whole to have
+a clear idea of the status of a project at any time.
+
+Adoption Process
+================
+
+An external project will exist in one of three phases of its lifecycle -
+adoption, where it's being discussed for inclusion in the Django project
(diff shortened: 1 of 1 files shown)


draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+    manner.
+
+Shepherd
+    The **Shepherd** is the Core Developer who will be the primary point of
+    contact for the project with the Core Team in Django, who will liase with
+    the Technical Board for the final vote, and who will assist in moving and
+    running the project under official Django ownership and infrastructure.
+    They can also be part of the Maintenance Team.
+
+The maintenance team for the adopted project may be different for that of the
```
phildini: I think you meant `may be different from that` for => from

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+or Django), and make sure all opinions are heard.
+
+Of particular note should be the presence of good documentation for the
+project; without this, it may be hard to discuss what it is or means. Projects
+with no documentation should likely not be considered for discussion before
+at least some documentation is written to anchor the discussion.
+
+The Shepherd should call an end to discussions after a reasonable time period;
+there is no requirement to wait until all discussions have "finished" before
+moving on (as this may take a very long time and needlessly long-winded
```
phildini: I would cut `needlessly long-winded discussions` -- it's weirdly tensed with the rest of the phrase, and also itself somewhat needless since the thought is covered by `a very long time`

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+The Technical Board are the final authority for deciding on adopting a project
+or not. They may choose to rule on the project as a team, or they
+may designate one or more board members to review and decide.
+
+The Technical Board should consider if the project's adoption would benefit
+Django; if by adopting they are crowding out other, potentially superior
+solutions; if the maintenance team is sufficient to ensure the project will
+be maintained properly once adopted; if the adoption of the project would
+place undue stress on the existing core team; and if adopting the project
+projects the right image and message about what Django is/
```
phildini: I think you meant `.` instead of `/`

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+
+* The top-level README of the project should be updated to officially list the
+  Shepherd and Maintenance Team, as well as details about Django's security
+  policy.
+
+* References to the project should be added in the official Django
+  documentation where sensible, as well as other changes made to ensure it's
+  discoverable.
+
+Ongoing Maintenance
```
phildini: General question: is the thought that documentation for the adopted project will be hosted somewhere on djangoproject.org, or will it use it's own docs hosting?

I, for one, am in favor of it living on some form of Django domain.

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+should ideally have more than one member so that this does not put the
+project at risk of retirement.
+
+Maintainers or people with commit access on an official Django project do not
+have to be core Django memebers, nor do they become core members by taking
+those positions, but they should be very strongly considered as candidates for
+the Core Team if they are not already.
+
+Retiring Projects
+=================
```
phildini: Retiring doesn't seem to indicate what will happen to the repository?

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+=========
+
+The goal of this proposed process is to enable a system where projects can
+be adopted and become official Django projects, but making sure it's not
+done prematurely (and thus stifling alternative approaches), and that it
+doesn't tie development to the release cadence and style of Django core
+releases too directly.
+
+That said, the Django name has gained a level of trust and respect over the
+years from our community and users, and so to become and official Django
```
phildini: I think you meant `an` instead of `and`.

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+Initial discussions should be held on mailing lists and other venues to
+solicit feedback from the community and work out if there is rough agreement
+that the project is a good thing for Django to adopt, particularly focusing
+on any alternative approaches to the same problem and the relative merits
+of them, including code design, scalability, alignment with existing Django
+design and philosophy, and having an available development and maintenance team.
+
+Notably, it does not need to be a maintainer of the existing project that
+starts this process, but a proposed project will need an implementation team
+willing to work with it once it is adopted to make it past the next step.
```
jacobian: But the maintainer(s) of the project probably need to be okay with us "adopting" it, right? We probably don't want to be in the business of adopting an actively maintained project over its maintainers objections, right?

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+The discussion is not expected to reach a consensus, though if it does that
+makes the review phase much easier; instead, it is meant for the community to
+discuss the pros and cons of adoption (the cons likely being alternative
+approaches to the same problem, or concerns that adoption will harm the project
+or Django), and make sure all opinions are heard.
+
+Of particular note should be the presence of good documentation for the
+project; without this, it may be hard to discuss what it is or means. Projects
+with no documentation should likely not be considered for discussion before
+at least some documentation is written to anchor the discussion.
```
jacobian: ❤️ 📖 💯

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,283 @@
...
+
+Once a project has been discussed and the discussion collated by the
+Maintenance Team and the Shepherd, it is moved onto review and decision by
+the Technical Board. The Shepherd will submit the project, the list of people
+signed up for the Maintenance Team, and the collated arguments to the
+Technical Board for decision.
+
+The Technical Board are the final authority for deciding on adopting a project
+or not. They may choose to rule on the project as a team, or they
+may designate one or more board members to review and decide.
```
jacobian: While I agree with what you've said above about consensus not being required, it's probably a good idea for the TB not to accept a project if it's super-controversial, right? That is, I'd suggest here that the TB's ruling should try to be roughly in-line with the majority PoV of the community. If a majority PoV doesn't exist, that might be a good reason for the TB to tread carefully.

draft/0007-official-projects.rst
jacobian: Given how great RTD is, I suggest we don't try to re-invent that wheel. Adding links to our documentation is probably good enough, IMO.

draft/0007-official-projects.rst
andrewgodwin: Done.

draft/0007-official-projects.rst
andrewgodwin: I've fleshed this out a bit. I would suggest we use RTD with a custom subdomain, maybe?

draft/0007-official-projects.rst
andrewgodwin: I will make that clearer; I would trust the TB not to make a controversial decision, but I'll highlight that they should err on the side of denial in those cases. Yes is forever, no is temporary.

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,298 @@
...
+adoption, where it's being discussed for inclusion in the Django project
+overall, maintenance, where it's part of the project and releasing new
+versions and applying bugfixes, and deprecation, where it's either lost all
+maintainers or is no longer needed, and is gently removed from the overall
+project.
+
+Adopting a project to be an official Django project consists of a number
+of phases, which are:
+
+1. `Pre-proposal`_ — a project looks like it may be worth adopting into Django
```
MarkusH: Use `--` instead of an en-dash, please. It's a renderer's task to show that correctly

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,298 @@
...
+Forming a team
+--------------
+
+Once the project looks like it might be a good candidate for an official
+Django project, a team must be formed to look after the adoption process,
+including:
+
+Maintenance Team
+    One or more people who are committed to maintaining the project once it
+    is part of Django. A project must have an active and willing maintenance
```
MarkusH: "... once it is under the Django umbrella", maybe? It's not "part of Django", IMO

andrewgodwin: I went with "once it's adopted" to try and be consistent.

draft/0007-official-projects.rst
andrewgodwin: Weird, DEP 001 has these in the source code too (I copied the formatting from there). I moved to colons instead.

MarkusH:
Can we add a requirement for a contributions guide to be available for a successful adoption (or at least during the ongoing maintenance)?

andrewgodwin:
@MarkusH Added some stuff about contributions for adoption and maintenance!

draft/0007-official-projects.rst
```diff
@@ -0,0 +1,307 @@
...
+policy to the core Django repository, nor are they subject to the same
+contribution patterns and guidelines as the core repository; how these work are
+up to the Maintenance Team.
+
+Maintainers are free to resign from their position at any time; the team
+should ideally have more than one member so that this does not put the
+project at risk of retirement.
+
+Maintainers or people with commit access on an official Django project do not
+have to be core Django memebers, nor do they become core members by taking
```
gvangool: *typo: "memebers" should "members"

MarkusH:
Thanks for the change @andrewgodwin :)

MarkusH:
Oh, and the DEP in general :smile:

MarkusH:
For end-of-live packages, do we also want a github org django-[attic,legacy,something] and move the repo there?

andrewgodwin:
@MarkusH We could, but my issue there is it means potentially breaking links from various other places. If we do though, we should call it -attic :)

andrewgodwin:
Since it looks like there's reasonable agreement on the draft, shall I merge it into the repo and take it out to the discussion phase?

collinanderson:
If you use github to move the repo, it redirects to the new location.

andrewgodwin:
I don't trust that feature to work and maintain a redirect forever, though.

MarkusH:
It's an official feature github offers and works until another repo with that name is created. See

* https://help.github.com/articles/renaming-a-repository/
* https://help.github.com/articles/transferring-a-repository-owned-by-your-organization/

andrewgodwin:
OK, I'll modify it to say we should move it under an organisation specifically for retired projects then, as the stronger the signal the better, and then merge it into the draft folder, as it seems like we all agree it's in good enough shape for a draft.

chicheng:
`Status:	Draft` not updated.

andrewgodwin:
Thanks, I have fixed that.

PR #25, opened 2016-07-15
Update the Status from Draft to Final in DEP 0007.

pelme:


mjtamlyn:
Looks like this was fixed in 9599c2819971ccd2aecf915d5b4f3628625e5eaa, thanks

PR #26, opened 2016-09-29
[WIP, do not merge] a new dependency policy

jacobian:
Python Packaging is Good Now.


glyph:
> This means that dependencies that require C extensions are probably not acceptable

What about projects which include manylinux1, windows, and mac wheels as part of their release process, and statically link their dependencies? In principle, it's totally possible to make C extension modules that do not require a C compiler to install nowadays.

This could create potential issues for FreeBSD users, for example, however, or people who use custom Pythons with weird compile-time options.  So I'm not saying this restriction should be removed, necessarily; just that it should be better-motivated.

jacobian:
@glyph FWIW I wrote that part before `manylinux1` was a thing :) This is right now a super-rough set of notes from when I first thought about it months ago, I turned it into a PR to sorta force myself to work on it more. I agree the C-ext part bears thinking through more clearly before setting a hard and fast rule.

draft/0007-dependancy-policy.rst
```diff
@@ -0,0 +1,67 @@
...
+Specification
+=============
+
+Guidelines for adding new dependencies
+--------------------------------------
+
+- External dependencies should be easy to install on all the platforms that Django supports (i.e. Linux/Mac/Windows, all supported Python versions including PyPy, etc). This means that dependencies that require C extensions are probably not acceptable.
+- stability
+- maintainability
+- "backup plan"
```
mjtamlyn: What happens if a dependency goes unmaintained definitely needs logging

alex:
Probably we should include a bit of guidance on how to do version pinning; I suspect we want it to look something like: "Only support libraries which have a decent backwards compatibility policy; put a reasonable minimum version and then no cap", but maybe we want to do semver style caps, `foo>=1.0,<2.0`

draft/0007-dependancy-policy.rst
```diff
@@ -0,0 +1,148 @@
+========================
```
adamchainz: filename should feature `dependency` not `dependancy` 😉

draft/0007-dependancy-policy.rst
```diff
@@ -0,0 +1,148 @@
...
+PyPI, including dependencies.
+
+.. [1] For more information on the early history of Python packaging, see
+       `this wonderful timeline <http://blog.startifact.com/posts/older/a-history-of-python-packaging.html>`_ that Martijn Faassen wrote up.
+
+However, Python packageing circa 2005 was just as rough around the edges as
+Django was at that time. ``virtualenv`` didn't exist yet, and system-wide
+installs were the norm (and just as problematic then as now). PyPI had
+occasional downtime, leading to frustration when trying to deploy to production.
+``easy_install`` failed in many corner casess. One of Django's early releases
```
akki: typo: `cases`.

draft/0007-dependancy-policy.rst
```diff
@@ -0,0 +1,148 @@
...
+    trying to create or consume Python libraries, it can be a tractable, even
+    pleasant experience.
+
+    I need to say this, because for a long time, Python’s packaging toolchain
+    was … problematic. It isn’t any more, but a lot of people still seem to
+    think that it is, so it’s time to set the record straight.
+
+Indeed. It's time for Django to let go of its decade-old suspicion of the
+packaging ecosystem. Python packaging is reliable and dependable, and it's time
+we took full advantage of features now available.
```
jezdez: This makes my heart be full ❤️

draft/0007-dependancy-policy.rst
```diff
@@ -0,0 +1,148 @@
...
+included in Django. For exmaple:
+
+- `passlib <https://pythonhosted.org/passlib/>`_ is a password hashing
+  library that implements a large variety of password hashing algorithms.
+  It's overlaps substantially with ``django.contrib.auth.hashers``, but
+  Django's version has fewer features than passlib.
+
+- Django implements its own internationalizing/localization framework, but many
+  developers feel `Babel <http://babel.pocoo.org/en/latest/>`_ is a superior
+  implementation.
```
jezdez: Huzzaaaah!

draft/0007-dependancy-policy.rst
```diff
@@ -19,30 +19,111 @@ Abstract
...
+
+That experience led to a deep suspicion of Python packaging tools among the
+Django core team, and a de-facto policy emerged of only requiring the lowest
+common denominator: instalation via direct download and ``python setup.py
+install``. This meant not using any of Python's packaging features developed
+since about 2002, including most notably dependecies. In 2016, Django's only
+dependencies are optional.
+
+However, a lot has changed in the last decade! In 2006, Django was pretty
+aweful: we had just `removed the magic
```
aaugustin: *awful

draft/0007-dependancy-policy.rst
aaugustin: Yes.

aaugustin:
I learnt things in this DEP :-)

shaib:
Continuing @alex's note, I think we need a checklist for the dependency-adding-mini-DEPs to answer:

- Does a specific minor version of Django depend on a specific version of the package? specific branch? If not, how do we guarantee backwards-compatibility for bugfix releases?
- How do we handle security/data-loss fixes? If relevant, can we get pre-notifications? If so, can we forward them?

In particular detail: Suppose Django 3.1.x depends on Frobnicate>=2.0,<3.0 as @alex suggested. At some point a vulnerability is discovered in Frobnicate<=2.3. Do we update the dependency to Frobnicate>=2.4,<3.0, or tell our users to keep tabs on Frobnicate on their own? If we do update, does such a change, in general, justify a Django security release?

In addition, we should consider LTS: How do we guarantee support for the term covered in our LTS policy? Should we limit the addition of dependencies to some part of our LTS cycle?

Also, I think we need to re-evaluate every dependency for every minor release. If some dependency's API becomes part of Django's API, then removing it becomes an issue; we should prepare for this -- one way would be forbidding ourselves to expose dependency APIs, but, frankly, I think that's not a workable limitation.

jacobian:
I think I've addressed the points raised by @alex @glyph @shaib. I'm going to make one final pass just for clarity and conciseness, but this is now done enough for me to merge.

draft/0007-dependency-policy.rst
```diff
@@ -0,0 +1,228 @@
...
+
+- Django vendors a version of `six <https://pythonhosted.org/six/>`_ (as
+  ``django.utils.six``). Instead of vendoring, we could use a dependency.
+
+- Much of Django's core HTTP/WSGI handling overlaps with utilities provided by
+  `Werkzeug <http://werkzeug.pocoo.org/>`_, the base underlying Flask and more.
+  If Django reimplemented its core HTTP/WSGI handling, we could share
+  maintenance burden with the Werkzeug/Flask maintainers while starting to
+  offer more opportunities for interoperability.
+
```
carljm: Add "test runner" as another example? Maybe don't need more examples...

draft/0007-dependency-policy.rst
```diff
@@ -0,0 +1,228 @@
...
+- **Stable** - Django's pretty solid at this point: it's well-tested,
+  production-proven, and relatively bug-free. All software has bugs, of course --
+  Django has 1,200 open issues as November 2016 -- but Django is, roughly
+  speaking, free of really critical bugs (crashers, data-loss, security issues,
+  etc). Dependencies need to be at a similar level of quality: code with serious
+  issues that wouldn't make it into Django shouldn't be accepted as a dependency,
+  either.
+
+- **Maintained** - if we discover bugs in a dependency, we need to be fairly 
+  confidant that they'll be fixed quickly.
```
carljm: confident, with an e (also next para)

draft/0007-dependency-policy.rst
```diff
@@ -0,0 +1,228 @@
...
+
+- **Takes security seriously** - we should be confidant that if we or our users
+  discover vulnerabilities in a dependency that the dependency authors will
+  respond to those vulnerabilities in coordination with us. This means they
+  should have a vulnerability disclosure policy, security-specific contacts,
+  and a history of taking vulnerabilities seriously.
+
+- **Works on all the same platforms as Django does** - Linux, Mac, Windows, 
+  and all supported Python versions (including PyPy). This probably means that
```
carljm: I see you doubled up on the "probably," just to be safe ;-)

draft/0007-dependency-policy.rst
```diff
@@ -0,0 +1,228 @@
...
+  wide a range of required versions as possible so that releases of Django
+  are de-coupled (as much as possible) from dependencies. Generally, we'll
+  want to specify dependencies as ``foo>=1.0,<2.0``, and be confidant that
+  point-releases of ``foo`` won't break Django. 
+
+Again, these are guidelines. At the end of the day, the criteria comes down to
+"would we include this code in Django?" The Tech Board has the final call.
+
+.. [1] Note the "probably" there. It is, in principle, possible to distribute 
+       C extensions in a way that no longer requires a complier -- platform-
```
carljm: compiler. Though a "complier" sounds like an interesting device...

draft/0007-dependency-policy.rst
```diff
@@ -0,0 +1,228 @@
...
+
+This DEP outlines Django's policy on external dependencies (e.g. other Python
+packages required to run Django). It supersedes Django's previous, unwritten
+policy of "no external dependencies allowed."
+
+In a nutshell, the policy is that adding a new external dependency should be
+treated similarly to adding a major new feature to Django: it requires a
+demonstration that the dependency is needed, and rough consensus among the
+community and core team that the chosen dependency is the correct one. If the
+dependency is controversial, it may require a DEP and a decision by the
```
carljm: This seems to suggest the idea of a DEP specifically for a dependency, which the Process section below suggests is not a thing

draft/0007-dependency-policy.rst
```diff
@@ -0,0 +1,228 @@
...
+
+    - Does the package meet the maturity bar laid out above? If there are 
+      any maturity risks -- for example, if the project only has a single 
+      maintainer -- that should be identified so we can do a cost/benefit
+      analysis.
+
+    - What version will we depend on? In general, we'd like to depend on a
+      wide range of versions (e.g. ``foo>1.0,<2.0``) so we can avoid tightly
+      coupling dependency releases to Django releases. But this may differ
+      from package to package, so the DEP should explain it closely.
```
shaib: The explanation here repeats a little of the above (under "backwards compatible")

PR #34, opened 2016-11-12
Several amendmants to DEP0201.

sjoerdjob:
- Being more specific on how to define a type converter.
- Mentioning the fact that parametrized type converters are a bit less
  specified (as of now) in contrast to the non-parametrized type converters,
  and might even be postponed till another DEP.
- Being more specific on reversing and resolving, and exception handling.
- Mention Django Channels and Django Hosts, and how it is not orthogonal right
  now.
- Added myself as author.

### DIFF
--- a/draft/0201-simplified-routing-syntax.rst
+++ b/draft/0201-simplified-routing-syntax.rst
@@ -155,12 +155,8 @@ Flask supports the `following converters <http://flask.pocoo.org/docs/0.11/quick
     Like ``int`` but for floating point values
 ``path``
     Like the default but also accepts slashes
-``any``
-    Matches one of the items provided
 ``uuid``
     Accepts UUID strings
-
-We might also consider including `a regex converter <http://stackoverflow.com/questions/5870188/does-flask-support-regular-expressions-in-its-url-routing>`_.
 
 Furthermore, an interface for implementing custom converters should exist. We
 could use the same API as Flask's ``BaseConverter`` for this purpose. The
@@ -210,30 +206,6 @@ The method ``to_url`` will always be called, no matter the type of ``value``.
 In particular, it will be called even when ``value`` is a string. This allows
 one to implement---for instance---a ``base64`` converter or a converter that
 works wth signed values as handled by ``django.core.signing.TimestampSigner``.
-
-Parametrized type converters
-----------------------------
-
-*The following behaviour is a bit less specified, and may even be postponed
-until a later date or another DEP*
-
-The ``any`` type converter as specified above would be useable as
-
-.. code-block:: python
-
-    path('/post/<any(int, string):id_or_slug>/', views.post_by_id_or_slug),
-
-This is to be implemented using parameter support. Likewise, there could be
-arguments to specify constraints, as follows
-
-.. code-block:: python
-
-    path('/user/<string(min_length=5, max_length=100):username>/', views.user)
-
-Implementation and parsing-wise this is a lot more unspecifiedd. One idea would
-be to supply the part between the ``<`` and the ``:`` to ``eval`` within a
-limited scope which only contains the registered converters (which is fine,
-*if* the ``path`` function can be safely assumed not to handle any user input).
 
 Adding type conversion to the existing system
 ---------------------------------------------
--- a/draft/0201-simplified-routing-syntax.rst
+++ b/draft/0201-simplified-routing-syntax.rst
@@ -171,17 +171,82 @@ registration of custom converters could be handled as a Django setting,
 Failure to perform a type conversion against a captured string should result in
 an ``Http404`` exception being raised.
 
+Definining type conversions
+---------------------------
+
+A converter is an object with three attributes/methods.
+
+``regex``
+    The pattern use in place of the type-specifier.
+``to_python``
+    How to convert the string from the URL to a Python object.
+``to_url``
+    How to convert the Python object back to something suitable in a URL.
+
+For instance, a converter for handling with the ``int`` parameter can be
+defined as follows.
+
+.. code-block:: python
+
+    class IntConverter(object):
+        regex = '-?[0-9]+'
+
+        def to_python(self, value):
+            return int(value)
+
+        def to_url(self, value):
+            return str(value)
+
+Here, ``to_python`` is going to be called as part of ``resolve`` while
+``to_url`` will be called during ``reverse``.
+
+If ``to_python`` raises a ``ValueError``, it will be interpreted as if the
+given path does not match the URL, and resolving will continue. This gives the
+ability to deal with cases where the validity of the content can not easily or
+fully be described using a regular expression alone. No other exceptions are
+caught.
+
+The method ``to_url`` will always be called, no matter the type of ``value``.
+In particular, it will be called even when ``value`` is a string. This allows
+one to implement---for instance---a ``base64`` converter or a converter that
+works wth signed values as handled by ``django.core.signing.TimestampSigner``.
+
+Parametrized type converters
+----------------------------
+
+*The following behaviour is a bit less specified, and may even be postponed
+until a later date or another DEP*
+
+The ``any`` type converter as specified above would be useable as
+
+.. code-block:: python
+
+    path('/post/<any(int, string):id_or_slug>/', views.post_by_id_or_slug),
+
+This is to be implemented using parameter support. Likewise, there could be
+arguments to specify constraints, as follows
+
+.. code-block:: python
+
+    path('/user/<string(min_length=5, max_length=100):username>/', views.user)
+
+Implementation and parsing-wise this is a lot more unspecifiedd. One idea would
+be to supply the part between the ``<`` and the ``:`` to ``eval`` within a
+limited scope which only contains the registered converters (which is fine,
+*if* the ``path`` function can be safely assumed not to handle any user input).
+
 Adding type conversion to the existing system
 ---------------------------------------------
 
 Adding a new URL syntax is easy enough, as they can be mapped onto the existing
 Regex syntax. The more involved piece of work would be providing for type
-conversion with the existing regex system. The type conversion functionality
-would need to support both named and unnamed capture groups.
+conversion with the existing regex system. It is our proposal that the type
+conversion (at first) only works for named capture groups. This because the
+``path`` function only builds named capture groups.
 
 One option could be:
 
-* Add a new ``converters`` argument to the ``url`` argument.
+* Add a new ``converters`` argument to the ``url`` function.
 * The value can either be a list/tuple, in which case its elements are mapped
   onto the capture groups by position, or a dict, in which case its elements
   are mapped onto the capture groups by name. (The former case is more general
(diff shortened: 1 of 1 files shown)


sjoerdjob:
@tomchristie: I added some statements which mirror my believe as "it is our believe". Could you double-check the changes, and make sure you still agree with the DEP. If not, I'll modify it.

draft/0201-simplified-routing-syntax.rst
```diff
@@ -171,30 +171,104 @@ registration of custom converters could be handled as a Django setting,
...
+
+Type conversions and ``reverse``
+--------------------------------
 
-(An alternative might be to add separate ``converter_args`` and
-``converter_kwargs`` arguments.)
+To support the ``reverse`` method on ``path``-based routes, the type converters
+will have to supply a ``to_url`` method which does the reversing. There will be
+no support for passing ``converter.to_url(value)`` to ``reverse``, because some
+``to_url`` functions might actually have text as input.
```
tomchristie: I don't understand this sentence?

draft/0201-simplified-routing-syntax.rst
```diff
@@ -171,30 +171,104 @@ registration of custom converters could be handled as a Django setting,
...
 
-We would also need to support the reverse side of type conversion. Ensure that
-reverse can be called with typed arguments as well as string literals.
+As an implementation detail, the plan is to call ``converter.to_url`` instead
+of ``force_text`` in ``_reverse_with_prefix``. The downside is that the
+conversion now has to happen inside a loop, instead of only once, which might
+have performance drawbacks.
```
tomchristie: This isn't clear to me either?

draft/0201-simplified-routing-syntax.rst
```diff
@@ -171,30 +171,104 @@ registration of custom converters could be handled as a Django setting,
...
+arguments to specify constraints, as follows
+
+.. code-block:: python
+
+    path('/user/<string(min_length=5, max_length=100):username>/', views.user)
+
+Implementation and parsing-wise this is a lot more unspecifiedd. One idea would
+be to supply the part between the ``<`` and the ``:`` to ``eval`` within a
+limited scope which only contains the registered converters (which is fine,
+*if* the ``path`` function can be safely assumed not to handle any user input).
```
tomchristie: I'd suggest that we leave the option of parameterized type convertors completely to one side for now. Let's get everything else working, documented, etc. and *then* consider it.

draft/0201-simplified-routing-syntax.rst
```diff
@@ -171,30 +171,104 @@ registration of custom converters could be handled as a Django setting,
...
-would need to support both named and unnamed capture groups.
+conversion with the existing regex system. It is our proposal that the type
+conversion (at first) only works for named capture groups. This because the
+``path`` function only builds named capture groups.
 
 One option could be:
 
-* Add a new ``converters`` argument to the ``url`` argument.
+* Add a new ``converters`` argument to the ``url`` function.
 * The value can either be a list/tuple, in which case its elements are mapped
```
tomchristie: If `convertors` can only be applied to named arguments then we should keep things simple and only support the dict style.

draft/0201-simplified-routing-syntax.rst
```diff
@@ -171,30 +171,104 @@ registration of custom converters could be handled as a Django setting,
...
 Adding type conversion to the existing system
 ---------------------------------------------
 
 Adding a new URL syntax is easy enough, as they can be mapped onto the existing
 Regex syntax. The more involved piece of work would be providing for type
-conversion with the existing regex system. The type conversion functionality
-would need to support both named and unnamed capture groups.
+conversion with the existing regex system. It is our proposal that the type
+conversion (at first) only works for named capture groups. This because the
+``path`` function only builds named capture groups.
```
tomchristie: We should probably note that the `convertors` argument is intended to be private-but-stable API, rather than documented API.

draft/0201-simplified-routing-syntax.rst
```diff
@@ -269,6 +344,20 @@ The following independent tasks can be identified:
...
+`Django Hosts <http://django-hosts.readthedocs.io/en/latest/>`_ allows for
+routing based on the host aspect of a request. Django Channels has a message
+routing layer, which can inspect different aspects of the messages.
+
+While it would be a good idea to see if the routing layer can be augmented to
+remove the need for django-hosts and be useful for Channels, it is our opinion
+that these are orthogonal concerns. Due to the expected implementation burden
+to also support these concerns, it is our preference that this is to be
+reconsidered at a later point in time, as to not delay the progress on the
+simplified routing syntax.
```
tomchristie: Don't think this is strong enough. Eg. "Due to the expected implementation burden to also support these concerns, it is our preference that this is to be reconsidered at a later point in time, as to not delay the progress on the simplified routing syntax." isn't a great argument.

Re. channels - really we'd like to see channels come into line with *this* proposal, rather than the other way around. Right now they're using a `path=<regex>` syntax, which is awkward.

I don't see any concerns around django-hosts - the different path syntax doesn't make any difference one way or the other there (although it's worth noting as an aside that we should ensure that we aim for a werkzeug compatible style, and also that werkzeug *does* have a host-based routing implementation)

draft/0201-simplified-routing-syntax.rst
sjoerdjob: I'm indifferent with respect to this. One good reason is that by making it private, we can forgo documenting it, and we don't have to keep it as stable as we otherwise would have to.

draft/0201-simplified-routing-syntax.rst
sjoerdjob: Sentences (or better: which of the two sentences).

```
To support the ``reverse`` method on ``path``-based routes, the type converters +will have to supply a ``to_url`` method which does the reversing.
```

is about what is technically needed to make `reverse` work with typed URLs: a `Converter` class must provide a method `to_url` which is the inverse of `to_python`.

```
There will be no support for passing ``converter.to_url(value)`` to ``reverse``, because some ``to_url`` functions might actually have text as input.
```

is about what happens when an argument to `reverse` is a string: the `converter.to_url` will still be called.

draft/0201-simplified-routing-syntax.rst
sjoerdjob: W.r.t. channels: yes, it would be great if it would use the new syntax as well. However, from what I understand, the question is more about the routing on *different* aspects of the request/message, and not so much about the syntax of the routing argument for a specific aspect.

What I hoped to make clear was that: "No, we won't support matching on other aspects of the request just yet. We can do that later, but that is not part of this DEP." Reasoning (to me) is basically "I expect it to be even harder than just supporting a better syntax. We can still do it later.".

If Channels is going to support the syntax of this proposal, that would be a big win. But I think that's something that @andrewgodwin has to determine for Channels, or at least not something that we should concern ourselves with. The best we can do (I think) is making sure the `'foo/<int:bar>/'` to `('^foo\/(?P<bar>[0-9]+)\/$', {'bar': IntConverter()})` conversion is decoupled and in a stable place.

andrewgodwin: The main problem Channels has is that it's a generic "message_item=regex" format, and so there's no special handling for any particular message item (be that URL path, method name, sender email, Slack channel name, etc.)

The API I was considering to solve this in future was something like `path=url('^foo/<int:bar>/')`, where the `url` object would be exposing a common interface that gave back a regex.

tomchristie: Granted that the default of `string`, which doesn't include the `/` character *is* due to URL paths. However I don't think there's any other aspects that are specific to URL paths, or that would make it unsuitable for generic pattern matching.

I can see that it'd be very awkward to find a transition path to using a different style, tho.

shaib:
All in all, this looks great. There are two points which I don't see covered in the current DEP in a satisfactory way; maybe it's just me.

First, the type conversions are only present in the simplified syntax (of course), but simplified and regex paths can co-exist in a system. This makes it possible, at least in theory, for a single view function to serve paths of both kinds. Type conversion implies validation, and a function which takes both validated and unvalidated input can easily become a security issue. I think this needs to be at least mentioned in documentation, perhaps even some measure taken against it (e.g -- probably a bad idea -- some decorator marking the view as suitable for one kind of invocation only).

Second, reusable apps are going to support a version of Django without simplified URL syntax for quite some time, and even after those versions are EOL'd, it is unlikely for most existing `urls.py` files to be replaced. This means that, in order for new users to understand them, the documentation should still include, pretty early on, a path to understanding regex paths.

As a thought that should probably have been brought up earlier: I know there was a discussion of whether to go with Python annotation syntax (`<foo:int>`) and it was preferred to use Flask URL syntax (`<int:foo>`); I haven't quite seen the discussion that led to the dropping of parametrization, but it occurred to me that parametrization might work better with Django Template Language filter syntax (`<foo|int:4>` for year).

tomchristie:
> I think this needs to be at least mentioned in documentation, perhaps even some measure taken against it.

Not sure - that *seems* like an over-complication. I don't see any obvious way of dealing with that elegantly. You could equally well have URL paths with *differing* type conversions pointed at the same view (incidentally we might want to test against that), which would seem to be valid, and isn't a terribly different case. From my point of view it'd be a slightly odd, but valid configuration - worth noting as an aside, but not something to be prevented.

> in order for new users to understand them, the documentation should still include, pretty early on, a path to understanding regex paths.

Agreed.

> I haven't quite seen the discussion that led to the dropping of parametrization

There hasn't actually been a discussion around that, which was an oversight more than anything. I think our current direction should be to start with the minimal viable patch, and consider addressing parameterization as a secondary concern. That's not making any pre-judgements about what would or wouldn't get accepted by the team, but just about prioritizing the basics first. I do think staying with a werkzeug-compatible style would be an advantage, tho.

sjoerdjob:
> (shaib): First, the type conversions are only present in the simplified syntax (of course), but simplified and regex paths can co-exist in a system. This makes it possible, at least in theory, for a single view function to serve paths of both kinds. Type conversion implies validation, and a function which takes both validated and unvalidated input can easily become a security issue. I think this needs to be at least mentioned in documentation, perhaps even some measure taken against it (e.g -- probably a bad idea -- some decorator marking the view as suitable for one kind of invocation only).

On the whole spectrum of validation, "Can be converted to an instance of `Foo`" is of course more validation than "Looks like something that can be converted to an instance of `Bar`". But, since the matching happens on regular expressions, I don't think this is significant difference that we should claim that it might become a security risk.

I feel that (at this point) this is not something that should be done.

> (tomchristie): ... You could equally well have URL paths with differing type conversions pointed at the same view (incidentally we might want to test against that), ...

For now, I would not test against that. It's a valid configuration. If it seems really troublesome at a later stage, but I don't currently see how it could be significantly troublesome.

> (shaib): I haven't quite seen the discussion that led to the dropping of parametrization

I agree with Tom his phrasing here: "I think our current direction should be to start with the minimal viable patch, and consider addressing parameterization as a secondary concern".

It was in the first version of the DEP, but in the end I believed that it was a secondary concern, and can be added at a later time without any backwards compatibility concerns. By working on a minimal viable patch we can (hopefully) get it into Django later.

draft/0201-simplified-routing-syntax.rst
sjoerdjob: @tomchristie: Does the above help you understand the sentence better? If so, could you help find a phrasing that makes it clear by default?

draft/0201-simplified-routing-syntax.rst
sjoerdjob: @tomchristie: Does the above help you understand the sentence better? If so, could you help find a phrasing that makes it clear by default?

EmilStenstrom:
I have read through the comments here, and I don't see any outstanding issues that have not been answered. Is there anything preventing this PR from being merged?

aaugustin:
I'm merging this version. We can iterate in a later PR if needed.

PR #36, opened 2016-11-15
Readonly and Auto-Refresh Fields

ewjoachim:
This is an early **very early** WIP on the DEP draft. The PR is just there so it's easier to gather opinions and move on.


draft/database-computed-fields.rst
```diff
@@ -0,0 +1,310 @@
...
+
+Questions left open
+^^^^^^^^^^^^^^^^^^^
+
+(When these questions are answered, the associated discussion should be summarized in "Rationale")
+
+Backdoor
+""""""""
+
+We could imagine different systems that would allow actually including the readonly fields for ``INSERT`` or ``UPDATES``. Do we need one ? For the tests ?
```
mjtamlyn: Raw SQL is always a back door

ewjoachim: Yes. But it's a last restort. We might want to have something better to offer to testers.

But we could also consider that they will (might) want to either test against a real database or mock the database completely.

ewjoachim:
@mjtamlyn To answer your review question, my guess would be something along the lines of :

```python
autofield = models.SerialField(primary_key=True, auto_refresh="create")
```

```python
uuid = models.UUIDField(primary_key=True, auto_refresh="create")
```

(`auto_refresh="create"` would also be used (probably assumed) when db_default is set which would make its implementation quite easier)

ewjoachim:
(as a non English native speaker, I'm not 100% sure I understood "a blank slate AutoField", so just to be sure : I undestood this as "how would this api help defining something like an autofield without using the already impleted autofield". Am I right ?)

mjtamlyn:
(correct, that's what I meant)

Can you expand on the rationale for a `readonly` field? I'm unclear what use case it's solving. Is it to strongly prevent the user from writing to it, or just allow the user to not write to it without it being null?

shaib:
On Tuesday 15 November 2016 19:25:40 Marc Tamlyn wrote:
> 
> Can you expand on the rationale for a `readonly` field? I'm unclear what
> use case it's solving. Is it to strongly prevent the user from writing to
> it, or just allow the user to not write to it without it being null?

Database-computed fields. Any attempt to write to them is an error.

adamchainz:
See e.g. https://mariadb.com/kb/en/mariadb/virtual-computed-columns/ . I know they don't exist in Postgres :wink:

shaib:
@adamchainz @mjtamlyn of course they do. https://www.postgresql.org/docs/9.5/static/sql-createview.html (look for "Updatable views")

ewjoachim:
Well, readonly is meant for any field where it's either an error or it make no sense at all to write it from Django. Usually, this starts happening when you have denormalization in your database, so it's a usecase for usually bigger databases.

I'll take a concrete example (there might not be the need for that much details here, but I might use this on the DEP itself)

You have :
```python
class Author(models.Model):
    name = models.CharField(max_length=100)

class Books(models.Model):
    author = models.ForeignKey(Author)
    number_sold = models.IntegerField()
```
You have many authors and each of them have many books. It starts becoming tedious to count the number of books sold by each Author, and as this is something you need done quite often because the top selling authors are displayed in real time on your home page, you decide to add a column on the `Author` model :
```python
    total_books_sold = models.IntegerField()
```
And if you want to make sure the new field is always up to date, you can write a database trigger that wil update it everytime a Book is updated.

But the when you do :
```
author = Author.objects.get(name="Terry Pratchett")
author.name = "Sir " + author.name
author.save()
```
If there was an update of total_books_sold by the DB between your ``get`` and your `` save`` you have probably overwritten it with the old value, and it will now be wrong until the next time you recompute completely the value from scratch.

Now you don't want an exception when this happens, because you did nothing wrong. But you don't want the database to write this field.

I might add, you could also say that you just don't define the field in Django, but if you don so, you can't easily read from it.

What you may want to be warned about is if you do:
```
terry_pratchett.total_books_sold = 12
# or
Authors.objects.filter(name="Terry Pratchett").update(total_books_sold=12)
# or
Authors.objects.create(name="Terry Pratchett", total_books_sold=12)
```

I'll try to make these use cases clearer in the DEP :+1: (I may use the example above)

Thank you a lot for your comments and your time. I still need to work a lot on the "auto-refresh" part. (I could use help, I think)

mjtamlyn:
Ok, I'm sold. That definitely makes a lot of sense. There are certainly merits to allowing the database to update certain fields with triggers, and it also gets us a lot closer to a `View` implementation.

ewjoachim:
Hmm, so I'm a little lost here :
1. should I be making a DEP on Readonly and Auto-refresh which IMO will help many other features ?(I guess the answer to this one is yes)
2. should I also be explicitely laying the foundations of db_default, serial field, and all other linked features that have been tried recently ? e.g. prviding a tentative specification, or mentionning that it will be easier when this DEP lands
3. Or should I be specifying them all to (but I know with certainty I alone won't be able to implement them all, and I'm not sure I can, by myself, line up a team of people willing to do it) ?

jarshwah:
I mentioned on IRC that readonly fields sound like an instance of database defaults. I realise now that this isn't totally the case.

In the case of database defaults and serial fields etc, I think it's OK to mention them as out of scope, but easier to implement provided this DEP is implemented.

I think it's a good idea to combine readonly fields with database refresh in the same DEP. I don't think you need to go into a huge amount of detail regarding the refresh. I would probably call out that RETURNING would be a good optimisation for databases that support it, but RETURNING doesn't need to be implemented to land readonly fields. The reason for mentioning refresh in this DEP is so that readers are aware of the reasons you're providing refresh options (incurring extra queries), without an expectation of RETURNING at this time.

Does that sound OK?

If so, we can create a new ticket for supporting RETURNING if one doesn't already exist. I don't think RETURNING as an optimisation requires a DEP unless you're aware of particular tricky details that might make implementation difficult or non-obvious.

I am hoping some others chime in on the refresh requirements relating to this DEP though.

ewjoachim:
That definitely sounds OK, @jarshwah ! If this DEP can go through with the elements you suggested, it means that 95% on the implementation is already present in django/django#7515 which already implements a basic version of "auto-refresh". The "RETURNING" optimisation is pretty much already done in django/django#2149 and django/django#5904 . It might represent a little more work but I'm confident. Especially if the API is already prepared and we just have to do the implementation (plus this allows to do it separately for Postgres and Oracle, so that's easier.

I was thinking : knowing that it will be trivial to implement, is it worth making the API of readonly also be able to use values `"create"` and `"update"` like auto_refresh would ? (you can have triggers on `UPDATE` but not on `INSERT` or vice versa. The 2 apis would be consistent and that may widen the possible usecases ...?

I'll try to finish the draft shortly !

charettes:
I just wanted to chime in to say that I think using the existing _deferred field_ concept for dealing with fields that needs to be refreshed because they are computed on database side would make a lot sense.

Having readonly fields trigger an extra query on every save until `RETURNING` support is added (or for backend not supporting it) could be a deal breaker for a lot of users. Imagine using this kind of field to mimic the `auto_now` feature of `DateFieldField` on the database side and having to fetch the data every single save even if you are probably not going to use the computed value anyway.

As I mentioned on django/django#7244 which proposes to implement a similar refresh for saved expressions:

> Accessing deferred fields would still yield the database computed value on-demand and in the rare cases where multiple expression fields are `save()`d at the same time it would still be possible to use `refresh_from_db(fields=[...])` to retrieve all of them in a single query instead of one per deferred fields.

ewjoachim:
I think defer is a good idea in theory, but I don't know enough of how it's coded to know if it's applicable. I know that until not so long ago, it would require changing the class of the returned model to a Deferred Model, which would make thing really more complicated, but if memory serves, this was changed, so I need to dive into the code again to see if it's now a more feasible thing.

ewjoachim:
Hmm, it looks like the new DEFERRED value for deferred fields will make thing really super easy to use. Ok, then !

adamchainz:
Yes `DEFERRED` makes sense to me too

eukreign:
Another use case for `readonly` and `deferred` is related to the postgres `tsvector` type.

If you have a trigger updating this column in the database then it should be set to `readonly` in Django.

Since the contents of the `tsvector` field are an implementation detail of postgres full text search, there is no use for it inside of django other than for debugging or some other analysis. Usually you want to just `.filter()` on this field but never actually retrieve it from the database.

This is a great use case for using `readonly` and `deferred` together.

ewjoachim:
(BTW, a quick update on that : I've been caught up in December life usual craziness, and my personal to-do list is quite long, but I plan to get back on this later in December (probably after the 25th)

ewjoachim:
I'd like @wengole 's opinion on the matter, but from my point of view, this DEP is ready for submission. I'll start working on a PoC implementation to have a clearer idea on the implications.

draft/database-computed-fields.rst
```diff
@@ -0,0 +1,364 @@
...
+   :depth: 3
+   :local:
+
+Abstract
+========
+
+This DEP defines 2 new properties of Model Fields:
+
+- ``readonly`` keeps Django from trying to write the value of those Fields to the database, either at creation time, update time or both;
+- ``auto_refresh`` makes Django aware that, when creating and/or updating a Field, its value needs to be fetched from the database.
```
wengole: Should this be `when creating and/or updating an object, the value of this Field will need to be fetched from the database` ?

draft/database-computed-fields.rst
```diff
@@ -0,0 +1,364 @@
...
+Specification
+=============
+
+There are two distinct parts in this DEP: Readonly and Auto-Refresh
+
+Readonly
+--------
+
+Of the two features described in this DEP, readonly should be the simplest one.
+It boils down to the following: when a ``Field`` is marked readonly, Django will not write it in the database, ever.
```
wengole: `Django will not write its value to the database, ever`

draft/database-computed-fields.rst
```diff
@@ -0,0 +1,364 @@
...
+
+Warnings
+^^^^^^^^
+
+The authors of the DEP are not aware of a preference between ``warnings`` (as in ``import warning``) or ``logs`` (as in ``logging.warning()``) as far of Django's own codebase goes.
+
+Serialization
+^^^^^^^^^^^^^
+
+As of now, a use-case has not been presented demonstrating the need for either excluding readonly fields from serialization or including them for deserialization.
```
wengole: Just something to bear in mind, but Django REST Framework (and others) may have issue if a `readonly` field is stripped from deserialization and it happened to the the identifying field. I'm not certain, however, that DRF uses Django's serialization.

charettes: I confirm it doesn't.

wengole:
Sorry I'd been absent from all the discussion, but I didn't realise it had got this far until @ewjoachim's mention yesterday.

I see nothing wrong with this DEP from a technical standpoint, and in fact it is most of what @ewjoachim and I covered at the DUTH sprints.

draft/database-computed-fields.rst
ewjoachim: Yes

draft/database-computed-fields.rst
ewjoachim: Well, as long as we don't have a use-case for excluding these fields, I think it's safe to keep it the way it is...

ewjoachim:
@wengole I've push a commit for your remarks :)

wengole:
Those changes are good, is there anything further blocking this?

draft/database-computed-fields.rst
```diff
@@ -0,0 +1,364 @@
...
+``readonly`` will be implemented by modifying the SQL compilers to make them "forget" about readonly fields, specifically for ``INSERT`` and ``UPDATE`` queries.
+
+In the ``SQLInsertCompiler`` class, the ``as_sql`` method manipulates ``self.query.fields``. We would change the implementation so that it would ignore the readonly fields (those whose readonly value is either ``True`` or ``"create"``).
+
+In the ``SQLUpdateCompiler`` class, the ``as_sql`` method manipulates ``self.query.values`` which is a list of tuples whose first values are ``Field`` instances. We would change the implementation so that it would ignore the readonly fields (those whose readonly value is either ``True`` or ``"update"``).
+
+Warning for misuses
+^^^^^^^^^^^^^^^^^^^
+
+We would log a ``warning`` (``logging.getLogger("django.db.models.readonly").warn()``) when calling :
```
jarshwah: Is a warning for modifying a readonly field appropriate? A lot of people don't run with warnings enabled. That said, I wonder how users might test readonly fields without some (relatively) complicated database initialisation. Have you thought through this?

draft/database-computed-fields.rst
```diff
@@ -0,0 +1,364 @@
...
+
+Serialization
+^^^^^^^^^^^^^
+
+Readonly fields would be included in serialization but silently excluded from deserialization. No warning would be logged.
+
+Empty queries
+^^^^^^^^^^^^^
+
+If a query becomes empty because all its fields are actually readonly, it should not be fired. This is already the case in the compilers, but this behaviour will be kept.
```
jarshwah: fired -> executed

draft/database-computed-fields.rst
```diff
@@ -0,0 +1,364 @@
...
+
+Auto-Refresh
+------------
+
+This part describes the Auto-refresh feature that will specify how, when field values are generated by the database, Django will fetch them and put them on their instances after a call to ``.save()``, ``Model.objects.create()``, and such.
+
+This will be done in the same query for the databases that support the ``RETURNING`` keyword. For other databases, it will be done by marking the field as ``deferred``. This way, the first time users try to access the value of a readonly field, it will be fetched from the database. Users may do an ``instance.refresh_from_db(readonly=True)`` to fetch them all with a single query.
```
jarshwah: Will the `refresh_from_db(readonly=True)` *only* refresh read only fields?

draft/database-computed-fields.rst
```diff
@@ -0,0 +1,364 @@
...
+^^^
+
+The proposed API would be:
+
+.. code-block:: python
+
+    class Book(models.Model):
+        refresh = models.IntegerField(auto_refresh=True)  # or "all"
+        refresh_at_creation = models.IntegerField(auto_refresh="create")
+        refresh_at_update = models.IntegerField(auto_refresh="update")
```
jarshwah: Is there a use case for `update` but not `create`?

draft/database-computed-fields.rst
wengole: I think that it would, yes. One use case being that it's possible you've set values on the object for writable fields, but want to get the DB provided values for the `readonly` fields without overriding your in-memory changes.

draft/database-computed-fields.rst
wengole: The only thing that springs to mind would be some kind of audit field that's only changed on an UPDATE and not on an INSERT. Can't think of a real world example, but having worked at a bank, I can imagine this being something that exists...

draft/database-computed-fields.rst
ewjoachim: Yes. I was wondering if "readonly" was clear enough for this usecase, but "only_readonly" feels really weird...

draft/database-computed-fields.rst
ewjoachim: Hmm, yes, maybe we could have just True/all, create and False. I tried to be consistent with the other API but really, there's no need, and it will complicate tests and all.

ewjoachim:
> My only concern at this point is how libraries are going to adapt to having to exclude readonly fields from their calculations. I think you should consider a section that discusses how the admin is going to handle readonly and refresh fields. Example: should textinputs be marked as readonly html attributes?

That's an interesting point. More than the Admin, I imagine the real problem would be ModelForms.

> The other thing is the string arguments to readonly and refresh ("create" and "update"). Did you consider separate kwargs or a constant defined somewhere? Something like models.UPDATE following in models.CASCADE footsteps. I'm not usually a fan of special strings, but a good constant name eludes me.

Yes, I was wondering. I'd like to avoir multiplicating kwargs (especially if we decide not to implement `auto_refresh="update"` as was discussed) but constants is definitely a good idea. If we share the same idea between the 2 features, we can share the constants. are models.CREATE, model.UPDATE, model.ALL acceptable ?

draft/database-computed-fields.rst
ewjoachim: The more I think about it, the more I feel it's YAGNI. I'll remove this part.

draft/database-computed-fields.rst
ewjoachim: After giving it a bit of thought, I'd say if we can't come up with an usecase, then, indeed, let's not do that. It will be trivial to add later if it comes up.

ewjoachim:
Regarding the ModelForm/Admin question, I think it's safe to say readonly fields are marked `editable=False` by default. This way, they won't appear and cause havoc in Admin / Model Forms. It's still possible to manually mark them as editable but then one has to deal with the consequences.

And finally, there's no penalty for trying to write a readonly field. Django will just ignore it (and issue a warning). So that should not be a problem for 3rd party apps.

ewjoachim:
(Following your suggestion I had a look at Models.CASCADE. It's actually not a constant (well not in the usual sense of it) but a function, so and so are all the other deletion choices... But yes, we can definitely have a real constant here :)

ewjoachim:
I'm less and less convinced that the "warning" part is either something we want, something we need, something we can afford in terms of performance (this one is probably negligible in most case but it's really at the core of some already overly optimised methods so we have to be cautious), and something we can possibly do reliably. I'd tend to actually skip this part (it can still be implemented independently).

Anyone's call on this ?

jarshwah:
RE: warnings - I'm not too bothered. Will the behaviour be to simply ignore the writes on readonly fields? Do you think usability will suffer as a consequence? Warnings are fairly useless for run time behaviour (rather than initialisation), but an error might be better on the set of the descriptor? Not sure how that would work when django loads the value though.

ewjoachim:
Ok... Spending one more day tinkering and chatting about that at DjangoConEurope, and I'm more and more going to the following conclusion:

- Auto-Refresh really is a subcase of Readonly. We could actually manage it with a single attribute or we could name the auto_refresh `readonly_auto_refresh` to make it clear
- When you mark your field readonly, by default it will be auto-refreshed except if you specifically ask that it isn't. This means 1 more query for DBs not supporting RETURNING and for everyone until RETURNING lands (but you **can** opt-out) (remember that the whole "readonly" is opt-in, so we're **not** talking about new queries for existing code)
- When you opt-out from the auto-refresh, then you don't have the extra query. Either your database supports RETURNING and it changes nothing, or it doesn't and it defers the fields.

ewjoachim:
Well, I'm really realizing we're trying to shovel use-cases all around in this DEP and that's why it's impossible to simplify it enough to get somewhere :(

So I guess:

1. We need to support internally `RETURNING` for any field and a fallback with a select query. This is the baseline of the implementation of Serial field, the implementation of database-level defauts (both are mentionned in this DEP) and also the proper implementation of readonly fields.

2. Once we've done that, and as part of a completely separate work, we need `readonly` on top of that and it's "trivial".

Doing it the other way around, really feels strange. Do you share that analysis or ..?

ewjoachim:
Oh, @jarshwah I just saw your answer (failed to refresh the page and GitHub did not show it).

We could try to make an error. If we do this, we have to make sure the error will always be there. I think it would be:
- Trivial for the case of `update(readonly=)` and .save(update_fields=["readonly"])
- More complex for the case of `instance.readonly = `, because the descriptor you mentioned whose set would trigger an error does not exist yet (if I'm not mistaken...) ? And then there's the problem you mentioned.
- And then if that's done, then `Model(readonly=)` will be covered too.

ewjoachim:
Ok, let's admit this PR will not be moving forward. So many things have changed, I guess even the DEP would need a complete do-over. Still, thank you :heart: See you around :)

PR #41, opened 2017-05-12
Finalize DEP 201

aaugustin:
This implements https://groups.google.com/forum/#!topic/django-developers/D44LSp0bPg8

I talked with Tom Christie on IRC and he approved my proposals.

### DIFF
--- a/accepted/0201-simplified-routing-syntax.rst
+++ b/accepted/0201-simplified-routing-syntax.rst
@@ -19,18 +19,16 @@ Abstract
 ========
 
 This DEP aims to introduce a simpler and more readable routing syntax to
-Django. Additionally the new syntax would support type coercion of URL
-parameters.
+Django. Additionally the new syntax supports type coercion of URL parameters.
 
-We would plan for this to become the new convention by default, but would do so
-in a backwards compatible manner, leaving the existing regex based syntax as an
-option.
+We plan for this to become the new convention by default, but in a backwards
+compatible manner, leaving the existing regex based syntax as an option.
 
 Motivation
 ==========
 
 Here's a section directly taken from Django's documentation on URL
-configuration...
+configuration:
 
 .. code-block:: python
 
@@ -43,7 +41,7 @@ configuration...
 
 There are two aspects to this that we'd like to improve on:
 
-* The Regex based URL syntax is unnecessarily verbose and complex for the vast
+* The regex-based URL syntax is unnecessarily verbose and complex for the vast
   majority of use-cases.
 * The existing URL resolver system does not handle typecasting, meaning that
   all URL parameters are treated as string literals.
@@ -56,11 +54,11 @@ provide for typecasting of the URL parameters that are passed to views.
 The existing syntax would remain available, and there would be no plans to
 place it on a deprecation path. Indeed, the underlying implementation for the
 typed URL syntax would actually be to use expand the typed URLs out into the
-existing Regex style, although this would largely remain an implementation
+existing regex style, although this would largely remain an implementation
 detail, rather than an exposed bit of API.
 
 The end result is that we would like to be able to present the following
-interface to our developers...
+interface to our developers:
 
 .. code-block:: python
 
@@ -71,9 +69,9 @@ interface to our developers...
         path('articles/<int:year>/<int:month>/<int:day>/', views.article_detail),
     ]
 
-The ``path()`` argument would also accept arguments without a converter prefix,
-in which case the converter would default to "string", accepting any text
-except a ``'/'``.
+The ``path()`` argument would also accept arguments without a converter
+prefix, in which case the converter would default to ``string``, accepting any
+text except a ``'/'``.
 
 For example:
 
@@ -81,21 +79,13 @@ For example:
 
     urlpatterns = [
         path('users/', views.user_list),
-        path('users/<id>/', views.user_detail),
+        path('users/<name>/', views.user_detail),
     ]
 
-For further background, please see the `"Challenge teaching Django to beginners: urls.py" <https://groups.google.com/forum/#!topic/django-developers/u6sQax3sjO4>`_ discussion group thread.
-
-Core vs Third-Party
-===================
-
-In our consideration this feature should be included in core Django rather than
-as a third-party app, because it adds significant value and readability.
-
-It is far more valuable when presented to the community as *the new standard*,
-rather than as an alternative style that can be bolted on. If presented as a
-third-party add-on then the expense of a codebase going against the standard
-URL convention will likely always prevent widespread uptake.
+For further background, please see the
+`Challenge teaching Django to beginners: urls.py
+<https://groups.google.com/forum/#!topic/django-developers/u6sQax3sjO4>`_
+discussion group thread.
 
 Specification
 =============
@@ -103,30 +93,30 @@ Specification
 Imports
 -------
 
-The naming for the import needs to be decided on. The existing URL configuration
-uses:
+The existing URL configuration uses:
 
 .. code-block:: python
 
     from django.conf.urls import url
 
-The naming question would be:
+The naming questions are:
 
-* What should the new style be called? Would we keep ``url``, or would we need
-  to introduce a different name to avoid confusion?
+* What should the new style be called? Would we keep ``url``, or would we
+  introduce a different name to avoid confusion?
 * Where should the new style be imported from?
 
-Our constraints here are that the existing naming makes sense, but we also need
-to ensure that we don't break backwards compatiblility.
+Our constraints here are that the existing naming makes sense, but we also
+need to minimize confusion, especially during the transition period, and to
+ensure that we don't break backwards compatiblility.
 
-Our proposal is that we should use a diffrent name and that the new style should
-be imported as...
+We propose to use a different name and to take this opportunity to simplify
+the import path:
 
 .. code-block:: python
 
     from django.urls import path
 
-A consistently named regex specific import would also be introduced...
+The regex version would be renamed consistently and made available as:
 
 .. code-block:: python
 
@@ -135,8 +125,8 @@ A consistently named regex specific import would also be introduced...
 The name ``path`` makes semantic sense here, because it actually does represent
 a URL component, rather than a complete URL.
 
-The existing import of ``from django.conf.urls import url`` would become a shim
-for the more explicit ``from django.urls import re_path``.
+The existing import of ``from django.conf.urls import url`` would become a
+shim for the more explicit ``from django.urls import re_path``.
 
 Given that it is currently used in 100% of Django projects, the smooth path
 for users would be to not deprecate ``django.conf.urls.url`` immediately, but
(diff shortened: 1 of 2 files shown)


aaugustin:
This PR moves the DEP to "Accepted", don't merge it until I can get confirmation from the Technical Board.

accepted/0201-simplified-routing-syntax.rst
```diff
@@ -56,124 +54,135 @@ provide for typecasting of the URL parameters that are passed to views.
...
+
+Furthermore, an interface for registering custom converters will be provided:
+
+.. code-block:: python
+
+    from django.urls import register_converter
+
+    register_converter(Converter, 'conv')
+
+Users are expected to call ``register_converter`` in their URLconf module, to
```
sjoerdjob: In that case, re-usable apps and all that, I'd expect to be able to register a converter in a namespaced manner:

```
register_converter(Converter, 'my_app_name.converter')
```

this to prevent potential clashes. Does that make sense? Otherwise, of course, one could still use it as

```
register_converter(Converter, 'my_app_name__converter')
```

sjoerdjob: This leaves us with some questions:

* How to deal with name clashes (they can be registered multiple times.)
* Suppose a converter is needed by multiple URLconf modules. How is a developer supposed to register the converter?

accepted/0201-simplified-routing-syntax.rst
```diff
@@ -56,124 +54,135 @@ provide for typecasting of the URL parameters that are passed to views.
...
+clear use case.
+
+Failure to perform a type conversion against a captured string is interpreted
+as if the given path does not match the URL.
 
 Definining type conversions
 ---------------------------
 
-A converter is an object with three attributes/methods.
+A converter is an object with three attributes or methods.
```
sjoerdjob: I'm thinking it might be better to use some wording to specify that `regex` is the attribute and `to_python`/`to_url` are the methods.

accepted/0201-simplified-routing-syntax.rst
```diff
@@ -239,41 +250,40 @@ of ``force_text`` in ``_reverse_with_prefix``. The downside is that the
...
-incorrectly.
+The following independent tasks can be identified:
 
-There is a decidedly edge-case deprecation that this would introduce in that
-existing projects that happen to *intentionally* include an unachored URL regex
-would raise a ``ConfigurationError`` when upgraded. However this is a loud and
-documentable error, with a simple resolution. (Change the import to
-``from django.urls import path_regex``.)
+* Implement several ``Converters``, and document the API.
+* Implement the ``converters`` argument. This adds the low-level API support
```
sjoerdjob: Not sure if this is an ordered list of tasks or not, but:

I'm thinking the `converters` argument should probably be implemented on `re_path`, and as such this task should probably be added a few lines down.

accepted/0201-simplified-routing-syntax.rst
```diff
@@ -239,41 +250,40 @@ of ``force_text`` in ``_reverse_with_prefix``. The downside is that the
...
+  for type coercion. Ensure that lookups perform type coercion, and
+  correspondingly, that calls to ``reverse`` work correctly with typed
+  arguments.
+* Add support for the new style ``path`` function, with an underlying
+  implementation based on the regex urls.
+* Add ``re_path``, with ``from django.conf.urls import url`` becoming a shim
+  for it.
+* Add support for registering custom converters, as defined in the Django
```
sjoerdjob: I think this is from before the switch to `register_converter`, correct?

accepted/0201-simplified-routing-syntax.rst
```diff
@@ -239,41 +250,40 @@ of ``force_text`` in ``_reverse_with_prefix``. The downside is that the
...
+  settings.
+* Document the new style URL configuration.
+* Update existing URL cases in the documentation throughout.
+* Update the tests throughout, updating to the new style wherever possible.
```
sjoerdjob: I'm not quite sure on this task: I could not find explicit tests on `url` and `resolve`. For `reverse` I could find the `urlpattern_reverse` test suite.

If we switch all the existing tests, we would add extra test coverage for `path`, yes. But we'll be decreasing coverage for `re_path`, and we should make sure to maintain enough tests as to make sure they don't break during the deprecation period, as well as making sure `re_path` stays tested.

accepted/0201-simplified-routing-syntax.rst
```diff
@@ -283,38 +293,39 @@ They should be implemented as a ``TypedURLPattern`` that subclasses
...
-Documentation
-=============
+Rationale
+=========
 
-The new style syntax would present a cleaner interface to developers. It would
-be beneficial for us to introduce the newer syntax as the primary style, with
-the existing regex style as a secondary option.
+Core vs Third-Party
+-------------------
```
sjoerdjob: As an extra argumentation (weak, I will admit) is that I actually tried implementing the `reverse` method for the case of stand-alone, and I just could not make that work. So we'd still have to add hooks to Django, and given the value-add for the functionality vs the value-add for the hooks, it clearly makes sense to just add the feature itself.

An argument for just adding the necessary hooks would be that we could have multiple implementing packages, and figure out which is the most popular before actually committing to merging into Django. However, seeing as we're mostly in agreement on this DEP, this does not seem necessary as a pre-step.

accepted/0201-simplified-routing-syntax.rst
aaugustin: I don't expect projects to have vast numbers of custom converters. I can't see how you'd have more than 10, let alone 20 or 30. So this question seems rather theoretical to me.

I'm not a fan of namespacing because it will affect the readability of URL patterns e.g. `/user/<magic_auth.id_conv:user>/property/<magic_auth.id_conv:property>/`.

For reusable apps, I think I'd just suggest to declare converters in each `path()` call that needs them — IIRC you (or someone else?) mentioned `path(..., converters={'name': Converter})` — and make sure this has precedence over globally-declared converters in case of a clash.

aaugustin: Again, I'm not too concerned with name clashes. I'd be surprised if they were an issue in practice, and if it is, we can make recommendations to avoid them.

Converters only need to be registered in the root URLconf when the URLconf is split across multiple modules.

If a project has multiple settings modules with independent URLconf modules (this happens), and there are enough converters to make it inconvenient to copy-paste their registration in each root URLconf (>10 converters and >10 root URLconfs seems rather unlikely), then it's possible register them in the `AppConfig.ready()` method of a shared application for example, or simply in a module that each URLconf imports.

accepted/0201-simplified-routing-syntax.rst
aaugustin: OK, will do.

accepted/0201-simplified-routing-syntax.rst
aaugustin: I didn't edit this section. It's more a checklist than an ordered list and it may not be complete.

accepted/0201-simplified-routing-syntax.rst
aaugustin: Yes I need to update this sentence.

accepted/0201-simplified-routing-syntax.rst
aaugustin: It's possible that url and resolve aren't tested specifically, but only as a side effect of other tests.

It would be very nice to add a new set of tests exercising specifically `path` and `path_re`. Then it doesn't matter what the rest of the test suite uses.

accepted/0201-simplified-routing-syntax.rst
aaugustin: Rough consensus and working code :-)

aaugustin:
Regarding the matter of declaring converters in settings, the arguments are the following.

Settings a big ball of global state that can affect arbitrary parts of the framework. It's hard to know how much depends on a particular setting. Many core devs think Django would be better off with fewer global settings. For this reason, the bar for adding a new setting is very high.

Over time we realized it's a better design if each component encapsulates its own configuration, for example, like the template engines do since 1.8. Registered converters are still a piece of unique global state — you won't be able to run two URLconfs within the same Python process with different converters — so we'll have the same drawbacks as global settings for testing, etc. but at least the state remains within the django.urls module, instead of being moved into the settings module.

I also find it preferrable to colocate the registration of new converters with the URLconf, because that reduces the distance between declaration and use.

aaugustin:
I'm happy to have that discussion on django-developers if that helps.

Be aware that arguing for adding a new setting is an uphill battle ;-)

aaugustin:
@sjoerdjob: thanks for all your comments. I just pushed a new commit to take them into account.

You may feel that I'm not getting to the bottom of things w.r.t `register_converter`. In fact I'm trying to minimize the API surface by not solving problems that may not exist.

It's quite likely that we'll make some additions once the new API has seen some real world use. Until then, keeping it minimal puts us in the best position to adapt as needed.

Please let me know if I can submit that version to the technical board.

sjoerdjob:
Yes, your explanation with respect to settings & `register_converter` has greatly helped.

sjoerdjob:
> Please let me know if I can submit that version to the technical board.

As far as I'm concerned: yes. I see nothing more left to add/remove.

accepted/0201-simplified-routing-syntax.rst
```diff
@@ -239,41 +263,42 @@ of ``force_text`` in ``_reverse_with_prefix``. The downside is that the
...
+  for type coercion. Ensure that lookups perform type coercion, and
+  correspondingly, that calls to ``reverse`` work correctly with typed
+  arguments.
+* Add support for registering custom converters.
+* Add dedicated tests for both the old and new style, so there's adequate
+  coverage. Currently only URL reversing is tested. URL resolution is only
+  exercised as a side effect of unrelated tests.
+* Document the new style URL configuration.
+* Update existing URL cases in the documentation throughout.
+* Update the tests throughout, updating to the new style wherever possible.
```
mjtamlyn: I'd like to add an additional (optional?) task here to make sure that customised 404 pages and resolving errors are tidied up. It would be nice if this showed the actual style used by the original code.

Also, Django has some URLs of its own (for example in admin). I don't think those can be converted as they use positional args (😡) but we should check if there are any others which need doing.

aaugustin: Thanks. I'll integrate your feedback and merge the PR ASAP.

aaugustin:
@sjoerdjob The technical board validates this version. I just need to find a few minutes to integrate Marc's latest comment before merging it. You can safely proceed with the implementation. Thanks!

accepted/0201-simplified-routing-syntax.rst
sjoerdjob: @mjtamlyn: to verify: do you mean the debug 404 error pages for resolving errors instead?

Also, regarding (for example) the Django admin URLs: is there any (implicit) backwards-compatibility guarantee on these?

mjtamlyn: I personally have a whole bunch of places where I reverse those URLs, and they are documented, so there is a backwards compat requirement. They also smell, really badly, so there may be a way to update them. Not as part of this work necessarily though!

The 404 thing I was talking about was the "default" debug 404 page which starts:
```
Using the URLconf defined in photocrowdsite.urls, Django tried these URL patterns, in this order:
^$ [name='home']
…
```

PR #62, opened 2019-09-12
Synchronize DEP folders with their status metadata.

ewjoachim:
A few DEPs have a status that says "Draft" in the header but are not
Draft anymore. This PR just fixes this.


ewjoachim:
I guess it's because these documents would be exepcted to be displayed not on GitHub ? But if it's not the case today, and not the case in the foreseeable future, we could remove the status and the 2 date fields that are duplicated info.

PR #63, opened 2019-09-13
Remove duplicate word "design" in Rationale section

mblayman:
I was using the DEP template as the basis of a proposal process at my job and found this extra word in the Rationale section.

Thanks for putting together this proposal process. I think that it is very well thought out.

I'm not sure what's going on with the final line in this PR. I used the editor on the GitHub website, and the editor did that. Functionally, it looks the same to me.

### DIFF
--- a/template.rst
+++ b/template.rst
@@ -75,7 +75,7 @@ Rationale
 =========
 
 This section should flesh out out the specification by describing what motivated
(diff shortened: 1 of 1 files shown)


MarkusH:
Thanks for catching this!

PR #84, opened 2023-12-19
DEP 84: Rejuvenate form media

thibaudcolas:
_Draft, not ready for review yet_

[Read the DEP as HTML](https://github.com/thibaudcolas/deps/blob/rejuvenating-form-media/draft/0000-rejuvenate-form-media.rst)

For context, see the forum thread: [Rejuvenating vs deprecating Form.Media](https://forum.djangoproject.com/t/rejuvenating-vs-deprecating-form-media/21285).

### DIFF
--- a/draft/0000-rejuvenate-form-media.rst
+++ b/draft/0000-rejuvenate-form-media.rst
@@ -1,0 +1,104 @@
+======================
+DEP XXXX: DEP template
+======================
+
+:DEP: XXXX
+:Author: Jacob Kaplan-Moss
+:Implementation Team: Jacob Kaplan-Moss
+:Shepherd: Andrew Godwin, Carl Meyer
+:Status: Draft
+:Type: Feature
+:Created: 2014-11-16
+:Last-Modified: 2014-11-18
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+This DEP provides a sample template for creating your own DEPs.  In conjunction
+with the content guidelines in `DEP 1 <https://github.com/django/deps/final/0001-dep-process.rst>`_,
+this should make it easy for you to conform your own DEPs to the format
+outlined below.
+
+Note: if you are reading this DEP via the web, you should first grab `the source
+of this DEP <https://raw.githubusercontent.com/django/deps/template.rst>`_ in
+order to complete the steps below.  **DO NOT USE THE HTML FILE AS YOUR
+TEMPLATE!**
+
+To get the source this (or any) DEP, look at the top of the Github page
+and click "raw".
+
+If you're unfamiliar with reStructuredText (the format required of DEPs),
+see these resources:
+
+* `A ReStructuredText Primer`__, a gentle introduction.
+* `Quick reStructuredText`__, a users' quick reference.
+* `reStructuredText Markup Specification`__, the final authority.
+
+__ http://docutils.sourceforge.net/docs/rst/quickstart.html
+__ http://docutils.sourceforge.net/docs/rst/quickref.html
+__ http://docutils.sourceforge.net/spec/rst/reStructuredText.html
+
+Once you've made a copy of this template, remove this abstract, fill out the
+metadata above and the sections below, then submit the DEP. Follow the
+guidelines in `DEP 1 <https://github.com/django/deps/final/0001-dep-process.rst>`_.
+
+Abstract
+========
+
+This should be a short (~200 word) description of the technical issue being
+addressed.
+
+This (and the above metadata) is the only section strictly required to submit a
+draft DEP; the following sections can be barebones and fleshed out as you work
+through the DEP process.
+
+Specification
+=============
+
+This section should contain a complete, detailed technical specification should
+describe the syntax and semantics of any new feature.  The specification should
+be detailed enough to allow implementation -- that is, developers other than the
+author should (given the right experience) be able to independently implement
+the feature, given only the DEP.
+
+Motivation
+==========
+
+This section should explain *why* this DEP is needed. The motivation is critical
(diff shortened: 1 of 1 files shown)


draft/0000-rejuvenate-form-media.rst
```diff
@@ -0,0 +1,95 @@
...
+- Dynamic module imports
+- async, defer scripts
+- CSPs via nonce attributes
+- integrity attribute
+- fetchpriority attribute
+- nomodule attribute
+- Arbitrary script attributes
+- Preloading / speculative loading
+- Resource ordering (see `capo.js <https://rviscomi.github.io/capo.js/>`_)
+- Web Components
```
matthiask, line 42: I think that's a very complete list. As I already said on the forum thread, I think that the frontend tooling space is still moving so quickly that it's hard to know which tools may or may not be useful in a few years time. When I think about Django and it's stability I have a horizon of several years always, with frontend tooling that's just not the case. 

So, I think there's a lot of value in trying to support CSP (that was one of the motivations for django-js-asset, see https://406.ch/writing/django-admin-apps-and-content-security-policy-compliance/ ), supporting additional arbitrary attributes on `<script>` tags etc. but that's it.

I'm not sure I understand the resource ordering thing. Django already supports relatively complex dependencies (at least since https://code.djangoproject.com/ticket/30153). Of course that doesn't do anything about the initial page loading performance, but then you're already in the area of bundlers again.

matthiask (commented):
I think introducing the `JS` class from django-js-asset into Django would be worthwile; probably a few parts of the admin should be refactored to take advantage of this? It's very stable, nothing has changed since the beginning of 2022.

claudep:
> I think introducing the [`JS` class from django-js-asset](https://github.com/matthiask/django-js-asset/blob/main/js_asset/js.py) into Django would be worthwile; probably a few parts of the admin should be refactored to take advantage of this? It's very stable, nothing has changed since the beginning of 2022.

Big +1. When I pushed for https://github.com/django/django/commit/4c76ffc2d6c77, it was my hope that at some point we could agree on some class-based JS/CSS objects (see the example in test_media.py).

draft/0000-rejuvenate-form-media.rst
thibaudcolas: From my perspective everything on that list has already proven its usefulness quite clearly! Additional attributes on `<script>` would be really useful but that feels to me like a rejuvenation of the Media API to bring it to 2014 standards rather than 2024.

thibaudcolas:
@matthiask would you be up for writing the "new attributes for `<script>` tags, and why we need them" side of the DEP? Then we can get things moving with this, and from my side I can focus on the research  I want to do for other requirements for merging of form widgets’ dependencies.

draft/0000-rejuvenate-form-media.rst
lb-: Should we also be adding anything about Django providing polyfills?

As in 

```

Opt in set of polyfills to support some/all of the above via pre-approved CDNs. Import maps, ESM, Web Components being the main ones. 
```

This is something the Rails approach did I believe, to allow for a 'simple' cross browser support approach where developers didn't have to think about what to include.

thibaudcolas:
Unfortunately I’ve decided I should stop working on this for now, so will close my draft :l @matthiask sorry for the start and stop! I still think this is very promising for Django, as a way to support (more) modern front-end development practices. But I’ve taken on new responsibilities as President of the DSF starting in December and I don’t see myself finding time in 2025 to work on substantive code contributions. I’m still happy to provide input occasionally but even for that I think I should focus on my commitments to Django’s accessibility team, so anything else will be "best effort" only.

---

With that said – yes please! I’d say import map merging is a great candidate for Django core’s `Media`. My rationale:

1. They seem pretty important to load ES modules at predictable paths while still allowing Django’s static files storage backends to do their thing.
2. Currently browsers only support one import map per page. Django’s current approach to leave UI / front-end to the packages ecosystem would work much better if it facilitates the merging, rather than having every package either teach its users how to manually merge, or create a thousand different APIs to do the merging.
3. They greatly simplify reusing JS between UI components, which is a big part of what `Media` is about

I think points 1. and 2. are pretty good reasons even without 3., but I tried to find good examples of "reuse between UI components" anyway.

- Rich text editors. In the case of Wagtail, there’s a built-in rich text editor, but a lot of people also want to use their own bespoke integration. It would be great if both flavors of the editor could load the same JS bundle rather than having it duplicated. The editor versions rarely change (6 releases in 6 years, only one with backwards-incompatible changes), so likelihood of compatibility between two integrations is high.
- The [Axe accessibility checker](https://github.com/dequelabs/axe-core). Also in Wagtail, we have it integrated into our "userbar" widget loaded on all pages. I’d also like to push our users towards a more developer-focused integration as a custom django-debug-toolbar panel. It’s a big dependency so crucial to dynamically load, and ideally only once.

## Other changes

In addition to the script attributes we already discussed, I think it’d also be good for `Media` to evolve by:

- Making it not just about forms and their widgets. I would guess this is 90-99% a documentation-only change. I don’t see any reason why it has to be recommended for forms only – see the Axe example above. For Wagtail we’ve done this and a bit more with the [laces](https://github.com/tbrlpld/laces) library by my colleague @tbrlpld.
- Supporting [rel=modulepreload](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/modulepreload). This is a natural next step once people start to use dynamic loading.

---

Sorry again I have to leave this there! I hope others will be able to make further progress.

matthiask:
Thanks for your thoughts and the support! I totally understand. If I find the time and energy I'll resurrect the DEP.

PR #86, opened 2024-02-09
Background workers

RealOrangeOne:
This is a proposal for adding a background workers interface to Django, to better facilitate developers (both working with Django and library maintainers) moving certain tasks from their web processes to separate background runners.

This work originally started in [Wagtail](https://wagtail.org) as [RFC 72](https://github.com/wagtail/rfcs/pull/72), but was found to be generic enough to also apply to Django, so @carltongibson suggested I adapt it to a DEP. See also my [more complete code examples](https://github.com/RealOrangeOne/django-core-tasks).

Got feedback? Feel free to leave a comment? Got _bigger_ feedback, I'm more than happy to chat on [Twitter](https://twitter.com/RealOrangeOne), [Mastodon](https://mastodon.theorangeone.net/@jake), or another channel if you already know me (so this PR doesn't get too messy).

[View rendered content](https://github.com/RealOrangeOne/deps/blob/background-workers/draft/0014-background-workers.rst)

### DIFF
--- a/accepted/0014-background-workers.rst
+++ b/accepted/0014-background-workers.rst
@@ -6,10 +6,10 @@ DEP 0014: Background workers
 :Author: Jake Howard
 :Implementation Team: Jake Howard
 :Shepherd: Carlton Gibson
-:Status: Draft
+:Status: Accepted
 :Type: Feature
 :Created: 2024-02-07
-:Last-Modified: 2024-04-19
+:Last-Modified: 2024-05-13
 
 .. contents:: Table of Contents
    :depth: 3
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -37,8 +37,8 @@ A backend will be a class which extends a Django-defined base class, and provide
    from datetime import datetime
    from typing import Callable, Dict, List
 
-   from django.contrib.tasks import BaseTask
-   from django.contrib.tasks.backends.base import BaseTaskBackend
+   from django.tasks import BaseTask
+   from django.tasks.backends.base import BaseTaskBackend
 
 
    class MyBackend(BaseTaskbackend):
@@ -98,7 +98,7 @@ A ``Task`` is used as a handle to the running task, and contains useful informat
    from datetime import datetime
    from typing import Any
 
-   from django.contrib.tasks import BaseTask, TaskStatus
+   from django.tasks import BaseTask, TaskStatus
 
    class MyBackendTask(BaseTask):
       id: str
@@ -154,7 +154,7 @@ Tasks can be queued using ``enqueue``, a proxy method which calls ``enqueue`` on
 
 .. code:: python
 
-   from django.contrib.tasks import enqueue
+   from django.tasks import enqueue
 
    def do_a_task(*args, **kwargs):
       pass
@@ -169,7 +169,7 @@ Similar methods are also available for ``defer``, ``aenqueue`` and ``adefer``. W
 
 .. code:: python
 
-   from django.contrib.tasks import tasks
+   from django.tasks import tasks
 
    def do_a_task(*args, **kwargs):
       pass
@@ -193,7 +193,7 @@ Tasks may also be "deferred" to run at a specific time in the future:
 
    from django.utils import timezone
    from datetime import timedelta
-   from django.contrib.tasks import defer
+   from django.tasks import defer
 
    task = defer(do_a_task, when=timezone.now() + timedelta(minutes=5))
 
@@ -213,7 +213,7 @@ Where the underlying task runner supports it, backends may also provide an ``asy
 
 .. code:: python
 
-   from django.contrib.tasks import aenqueue
+   from django.tasks import aenqueue
 
    await aenqueue(do_a_task)
 
@@ -224,7 +224,7 @@ Settings
 
    TASKS = {
       "default": {
-         "BACKEND": "django.contrib.tasks.backends.ImmediateBackend",
+         "BACKEND": "django.tasks.backends.ImmediateBackend",
          "OPTIONS": {}
       }
    }
@@ -260,7 +260,7 @@ The global task connection ``tasks`` is used to access the configured backends, 
 
 .. code:: python
 
-   from django.contrib.tasks import task
+   from django.tasks import task
 
    # Later...
    task = task.enqueue(do_a_thing)
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -108,7 +108,7 @@ A ``Task`` is used as a handle to the running task, and contains useful informat
 .. code:: python
 
    from datetime import datetime
-   from typing import Any
+   from typing import Any, Callable
 
    from django.tasks import BaseTask, TaskStatus
 
@@ -130,6 +130,18 @@ A ``Task`` is used as a handle to the running task, and contains useful informat
 
       raw: Any | None
       """Return the underlying runner's task handle"""
+
+      priority: int | None
+      """The priority of the task"""
+
+      func: Callable
+      """The task function"""
+
+      args: list
+      """The arguments to pass to the task function"""
+
+      kwargs: dict
+      """The keyword arguments to pass to the task function"""
 
       def __init__(self, **kwargs):
          """
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -62,7 +62,7 @@ A backend will be a class which extends a Django-defined base class, and provide
 
       def defer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
          """
-         Add a task function (or coroutine) to be completed at a specific time
+         Add a task function (or coroutine) to be completed at a specific (timezone-aware) time
          """
          ...
 
@@ -74,7 +74,7 @@ A backend will be a class which extends a Django-defined base class, and provide
 
       async def adefer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
          """
-         Add a task function (or coroutine) to be completed at a specific time
+         Add a task function (or coroutine) to be completed at a specific (timezone-aware) time
          """
          ...
 
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -100,7 +100,7 @@ A backend will be a class which extends a Django-defined base class, and provide
 
 If a backend doesn't support a particular scheduling mode, it simply does not define the method. Convenience methods ``supports_enqueue`` and ``supports_defer`` will be implemented by ``BaseTaskBackend``. Similarly, ``BaseTaskBackend`` will provide ``a``-prefixed stubs for ``enqueue``, ``defer`` and ``get_task`` wrapped with ``asgiref.sync_to_async``.
 
-``is_valid_task_function`` determines whether the provided function (or possibly coroutine) is valid for the backend. This can be used to prevent coroutines from being executed, or otherwise validate the callable. The default implementation will ensure the callable is globally importable.
+``is_valid_task_function`` determines whether the provided function (or possibly coroutine) is valid for the backend. This can be used to prevent coroutines from being executed, or otherwise validate the callable.
 
 Django will ship with 3 implementations:
 
@@ -187,6 +187,28 @@ A ``Task``'s ``status`` must be one of the follwing values (as defined by an ``e
 
 If a backend supports more than these statuses, it should compress them into one of these.
 
+Task functions
+--------------
+
+A task function is any globally-importable callable which can be used as the function for a task (ie passed into ``enqueue``).
+
+Before a task can be run, it must be marked:
+
+.. code:: python
+
+   from django.tasks import task
+
+   @task
+   def do_a_task(*args, **kwargs):
+      pass
+
+The decorator "marks" the task as being a valid function to be executed. This prevent arbitrary methods from being queued, potentially resulting in a security vulnerability (eg ``subprocess.run``).
+
+Tasks will be validated against the backend's ``is_valid_task_function`` before queueing. The default implementation will validate all generic assumptions:
+
+- Is the task function globally importable
+- Has the task function been marked
+
 Queueing tasks
 -------------
 
@@ -194,8 +216,9 @@ Tasks can be queued using ``enqueue``, a proxy method which calls ``enqueue`` on
 
 .. code:: python
 
-   from django.tasks import enqueue
+   from django.tasks import enqueue, task
 
+   @task
    def do_a_task(*args, **kwargs):
       pass
 
@@ -209,8 +232,9 @@ Similar methods are also available for ``defer``, ``aenqueue`` and ``adefer``. W
 
 .. code:: python
 
-   from django.tasks import tasks
+   from django.tasks import tasks, task
 
+   @task
    def do_a_task(*args, **kwargs):
       pass
 
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -257,6 +257,21 @@ Where the underlying task runner supports it, backends may also provide an ``asy
 
    await aenqueue(do_a_task)
 
+Similarly, a backend may support queueing an async task function:
+
+.. code:: python
+
+   from django.tasks import aenqueue, enqueue, task
+
+   @task
+   async def do_an_async_task():
+      pass
+
+   await aenqueue(do_an_async_task)
+
+   # Also works
+   enqueue(do_an_async_task)
+
 Settings
 ---------
 
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -142,6 +142,8 @@ A ``Task`` can only be created for globally-importable callables. The task will 
 
 ``@task`` may be used on functions or coroutines. It will be up to the backend implementor to determine whether coroutines are supported. In either case, the function must be globally importable.
 
+Task arguments must be JSON serializable, to avoid compatibility and versioning issues. Complex arguments should be converted to a format which is JSON-serializable.
+
 Task Results
 ------------
 
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -44,11 +44,11 @@ A backend will be a class which extends a Django-defined base class, and provide
    class MyBackend(BaseTaskbackend):
       task_class = Task
 
-      def __init__(self, options: Dict):
+      def __init__(self, settings_dict: Dict):
          """
          Any connections which need to be setup can be done here
          """
-         super().__init__(options)
+         super().__init__(settings_dict)
 
       @classmethod
       def is_valid_task(cls, task: Task) -> bool:
@@ -320,8 +320,6 @@ Settings
 
 
 ``QUEUES`` contains a list of valid queue names for the backend. If a task is queued to a queue which doesn't exist, an exception is raised. If omitted or empty, any name is valid.
-
-``OPTIONS`` is passed as-is to the backend's constructor. ``QUEUES`` is additionally passed to the constructor as the ``queues`` keyword argument.
 
 Motivation
 ==========
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -247,7 +247,7 @@ But what about *X*?
 
 The most obvious alternative to this DEP would be to standardise on a task implementation and vendor it in to Django. The Django ecosystem is already full of background worker libraries, eg Celery and RQ. Writing a production-ready task runner is a complex and nuanced undertaking, and discarding the work already done is a waste.
 
-This proposal doesn't seek to replace existing tools, nor add yet another option for developers to consider. The primary motivation is creating a shared API contract between worker libaries and developers. It does however provide a simple way to get started, with a solution suitable for most sizes of projects (``DatabaseBackend``).
+This proposal doesn't seek to replace existing tools, nor add yet another option for developers to consider. The primary motivation is creating a shared API contract between worker libaries and developers. It does however provide a simple way to get started, with a solution suitable for most sizes of projects (``DatabaseBackend``). Slowly increasing features, adding more built-in storage backends and a first-party task runner aren't out of the question for the future, but must be done with careful planning and consideration.
 
 Rationale
 =========
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -1,0 +1,299 @@
+=============================
+DEP XXXX: Background workers
+=============================
+
+:DEP: XXXX
+:Author: Jake Howard
+:Implementation Team: Jake Howard
+:Shepherd: Carlton Gibson
+:Status: Draft
+:Type: Feature
+:Created: 2024-02-07
+:Last-Modified: 2024-02-09
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+Abstract
+========
+
+Django doesn't have a first-party solution for long-running tasks, however the ecosystem is filled with incredibly popular frameworks, all of which interact with Django in slightly different ways. Other frameworks such as Laravel have background workers built-in, allowing them to push tasks into the background to be processed at a later date, without requiring the end user to wait for them to occur.
+
+Library maintainers must implement support for any possible task backend separately, should they wish to offload functionality to the background. This includes smaller libraries, but also larger meta-frameworks with their own package ecosystem such as `Wagtail <https://wagtail.org>`_.
+
+Specification
+=============
+
+The proposed implementation will be in the form of an application wide "task backend" interface. This backend will be what connects Django to the task runners with a single pattern. The task backend will provide an interface for either third-party libraries, or application developers to specify how tasks should be created and pushed into the background
+
+Backends
+--------
+
+A backend will be a class which extends a Django-defined base class, and provides the common interface between Django and the underlying task runner.
+
+.. code:: python
+
+   from datetime import datetime
+   from typing import Callable, Dict, List
+
+   from django.contrib.tasks import BaseTask
+   from django.contrib.tasks.backends.base import BaseTaskBackend
+
+
+   class MyBackend(BaseTaskbackend):
+      def __init__(self, options: Dict):
+         """
+         Any connections which need to be setup can be done here
+         """
+         super().__init__(options)
+
+      def enqueue(self, func: Callable, priority: int | None, args: List, kwargs: Dict) -> BaseTask:
+         """
+         Queue up a task function to be executed
+         """
+         ...
+
+      def defer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
+         """
+         Add a task to be completed at a specific time
+         """
+         ...
+
+      async def aenqueue(self, func: Callable, priority: int | None, args: List, kwargs: Dict) -> BaseTask:
+         """
+         Queue up a task function (or coroutine) to be executed
+         """
+         ...
+
+      async def adefer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
+         """
+         Add a task function (or coroutine) to be completed at a specific time
+         """
+         ...
+
+      def close(self) -> None:
+         """
+         Close any connections opened as part of the constructor
+         """
+         ...
+
+If a backend doesn't support a particular scheduling mode, it simply does not define the method. Convenience methods ``supports_async`` and ``supports_defer`` will be implemented by ``BaseTaskBackend``.
+
+Django will ship with 2 implementations:
+
+ImmediateBackend
+   This backend runs the tasks immediately, rather than offloading to a background process. This is useful both for a graceful transition towards background workers, but without impacting existing functionality.
+
+DatabaseBackend
+   This backend uses the Django ORM as a task store. This backend will support all features, and should be considered production-grade.
+
+Tasks
+-----
+
+A ``Task`` is used as a handle to the running task, and contains useful information the application may need when referencing the task.
+
+.. code:: python
+
+   from datetime import datetime
+   from typing import Any
+
+   from django.contrib.tasks import BaseTask, TaskStatus
+
+   class MyBackendTask(BaseTask):
+      id: str
+      """A unique identifier for the task"""
+
+      status: TaskStatus
+      """The status of the task"""
+
+      result: Any
+      """The return value from the task function"""
+
+      queued_at: datetime
+      """When the task was added to the queue"""
+
+      completed_at: datetime | None
+      """When the task was completed"""
+
+      raw: Any | None
+      """Return the underlying runner's task handle"""
+
+      def __init__(self, **kwargs):
+         """
+         Unpacking the raw response from the backend and storing it here for future use
+         """
+         super().__init__(**kwargs)
+
+      def refresh(self) -> None:
+         """
+         Reload the cached task data from the task store
+         """
+         ...
+
+      async def arefresh(self) -> None:
+         """
+         Reload the cached task data from the task store
+         """
+         ...
+
+
+A ``Task`` will cache its values, relying on the user calling ``refresh`` / ``arefresh`` to reload the values from the task store.
+
+To enable a ``Task`` to be backend-agnostic, statuses must include a set of known values. Additional fields may be added if the backend supports them, but these attributes must be supported:
+
+:New: The task has been created, but hasn't started running yet
+:Running: The task is currently running
+:Failed: The task failed
+:Complete: The task is complete, and the result is accessible
+
+Running tasks
+-------------
+
+Tasks can be queued using ``enqueue``, a proxy method which calls ``enqueue`` on the default task backend:
+
+.. code:: python
+
+   from django.contrib.tasks import enqueue
+
+   def do_a_task(*args, **kwargs):
+      pass
+
+   # Submit the task function to be run
+   task = enqueue(do_a_task)
+
+   # Optionally, provide arguments
+   task = enqueue(do_a_task, args=[], kwargs={})]
+
+Similar methods are also available for ``defer``, ``aenqueue`` and ``adefer``. When multiple task backends are configured, each can be obtained from a global ``tasks`` connection handler:
+
+.. code:: python
+
+   from django.contrib.tasks import tasks
+
+   def do_a_task(*args, **kwargs):
+      pass
+
+   # Submit the task function to be run
+   task = tasks["special"].enqueue(do_a_task)
+
+   # Optionally, provide arguments
+   task = tasks["special"].enqueue(do_a_task, args=[], kwargs={})
+
+When enqueueing tasks, ``args`` and ``kwargs`` are intentionally their own dedicated arguments to make the API simpler and backwards-compatible should other attributes be added in future.
+
+Here, ``do_a_task`` can either be a regular function or coroutine. It will be up to the backend implementor to determine whether coroutines are supported.
+
+Deferring tasks
+---------------
+
+Tasks may also be "deferred" to run at a specific time in the future:
+
+.. code:: python
+
+   from django.utils import timezone
+   from datetime import timedelta
+   from django.contrib.tasks import defer
+
+   task = defer(do_a_task, when=timezone.now() + timedelta(minutes=5))
+
+When scheduling a task, it may not be **exactly** that time a task is executed, however it should be accurate to within a few seconds. This will depend on the current state of the queue and task runners, and is out of the control of Django.
+
+Sending emails
+--------------
+
+One of the easiest and most common places that offloading work to the background can be performed is sending emails. Sending an email requires communicating with an external, potentially third-party service, which adds additional latency and risk to web requests. These can be easily offloaded to the background.
+
+Django will ship with an additional task-based SMTP email backend, configured identically to the existing SMTP backend. The other backends included with Django don't benefit from being moved to the background.
+
+Async tasks
+-----------
+
+Where the underlying task runner supports it, backends may also provide an ``async``-compatible interface for task running, using ``a``-prefixed methods:
+
+.. code:: python
+
+   from django.contrib.tasks import aenqueue
+
+   await aenqueue(do_a_task)
+
+Settings
+---------
+
+.. code:: python
+
+   TASKS = {
+      "default": {
+         "BACKEND": "django.contrib.tasks.backends.ImmediateBackend",
+         "OPTIONS": {}
+      }
+   }
+
+``OPTIONS`` is passed as-is to the backend's constructor.
+
+Motivation
+==========
+
+Having a first-party interface for background workers poses 2 main benefits:
+
+Firstly, it lowers the barrier to entry for offloading computation to the background. Currently, a user needs to research different worker technologies, follow their integration tutorial, and modify how their tasks are called. Instead, a developer simply needs to install the dependencies, and work out how to *run* the background worker. Similarly, a developer can start determining which actions should run in the background before implementing a true background worker, and avoid refactoring should the backend change over time.
+
+Secondly, it allows third-party libraries to offload some of their execution. Currently, library maintainers need to either accept their code will run inside the request-response lifecycle, or provide hooks for application developers to offload actions themselves. This can be particularly helpful when offloading certain expensive signals.
+
+One of the key benefits behind background workers is removing the requirement for the user to wait for tasks they don't need to, moving computation and complexity out of the request-response cycle, towards dedicated background worker processes. Moving certain actions to be run in the background not improves performance of web requests, but also allows those actions to run on specialised hardware, potentially scaled differently to the web servers. This presents an opportunity to greatly decrease the percieved execution time of certain common actions performed by Django projects.
+
+But what about *X*?
+-------------------
+
+The most obvious alternative to this DEP would be to standardise on a task implementation and vendor it in to Django. The Django ecosystem is already full of background worker libraries, eg Celery and RQ. Writing a production-ready task runner is a complex and nuanced undertaking, and discarding the work already done is a waste.
+
+This proposal doesn't seek to replace existing tools, nor add yet another option for developers to consider. The primary motivation is creating a shared API contract between worker libaries and developers. It does however provide a simple way to get started, with a solution suitable for most sizes of projects (``DatabaseBackend``).
+
+Rationale
+=========
+
+This proposed implementation specifically doesn't assume anything about the user's setup. This not only reduces the chances of Django conflicting with existing task systems a user may be using (eg Celery, RQ), but also allows it to work with almost any hosting environment a user might be using.
+
+This proposal started out as `Wagtail RFC 72 <https://github.com/wagtail/rfcs/pull/72>`_, as it was becoming clear a unified interface for background tasks was required, without imposing on a developer's decisions for how the tasks are executed. Wagtail is run in many different forms at many differnt scales, so it needed to be possible to allow developers to choose the backend they're comfortable with, in a way which Wagtail and its associated packages can execute tasks without assuming anything of the environment it's running in.
+
+The global task connection ``tasks`` is used to access the configured backends, with global versions of those methods available for the default backend. This contradicts the pattern already used for storage and caches. A "task" is already used in a number of places to refer to an executed task, so using it to refer to the default backend is confusing and may lead to it being overridden in the current scope:
+
+.. code:: python
+
+   from django.contrib.tasks import task
+
+   # Later...
+   task = task.enqueue(do_a_thing)
+
+   # Clearer
+   thing_task = task.enqueue(do_a_thing)
+
+Backwards Compatibility
+=======================
+
+So that library maintainers can use this integration without concern as to whether a Django project has configured background workers, the default configuration will use the ``ImmediateBackend``. Developers on older versions of Django but who need libraries which assume tasks are available can use the reference implementation.
+
+Reference Implementation
+========================
+
+The reference implementation is currently being developed alongside this DEP process. This implementation will serve both as an "early-access" demo to get initial feedback and start using the interface, as the basis for the integration with Django core, but also as a backport for users of supported Django versions prior to this work being released.
+
+Once code is available, it will be referenced here.
+
+Future iterations
+=================
+
+The field of background tasks is vast, and attempting to implement everything supported by existing tools in the first iteration is futile. The following functionality has been considered, and deemed explicitly out of scope of the first pass, but still worthy of future development:
+
+- Completion hooks, to run subsequent tasks automatically
+- Bulk queueing
+- Automated task retrying
+- A generic way of executing task runners. This will remain the repsonsibility of the underlying implementation, and the user to execute correctly.
+- Observability into task queues, including monitoring and reporting
+- Cron-based scheduling. For now, this can be achieved by triggering the same task once it finishes.
+
+Copyright
+=========
+
+This document has been placed in the public domain per the Creative Commons
+CC0 1.0 Universal license (http://creativecommons.org/publicdomain/zero/1.0/deed).
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -178,12 +178,14 @@ A ``Task`` is obtained either when scheduling a task function, or by calling ``g
 
 A ``Task`` will cache its values, relying on the user calling ``refresh`` / ``arefresh`` to reload the values from the task store.
 
-To enable a ``Task`` to be backend-agnostic, statuses must include a set of known values. Additional fields may be added if the backend supports them, but these attributes must be supported:
+A ``Task``'s ``status`` must be one of the follwing values (as defined by an ``enum``):
 
 :NEW: The task has been created, but hasn't started running yet
 :RUNNING: The task is currently running
 :FAILED: The task failed
 :COMPLETE: The task is complete, and the result is accessible
+
+If a backend supports more than these statuses, it should compress them into one of these.
 
 Queueing tasks
 -------------
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -57,7 +57,7 @@ A backend will be a class which extends a Django-defined base class, and provide
          """
          ...
 
-      def enqueue(self, task: Task, *, args: List, kwargs: Dict, priority: int | None = None, run_after: datetime | None = None) -> TaskResult:
+      def enqueue(self, task: Task, *, args: List, kwargs: Dict, priority: int | None = None, run_after: datetime | None = None, queue_name: str | None = None) -> TaskResult:
          """
          Queue up a task to be executed
          """
@@ -123,6 +123,9 @@ Backend implementors aren't required to implement their own ``Task``, but may fo
       kwargs: dict
       """The keyword arguments to pass to the task function"""
 
+      queue_name: str | None
+      """The name of the queue the task will run on """
+
 
 A ``Task`` is created by decorating a function with ``@task``:
 
@@ -168,6 +171,9 @@ Backend implementors aren't required to implement their own ``TaskResult``, but 
       status: TaskStatus
       """The status of the running task"""
 
+      queue_name: str | None
+      """The name of the queue the task will run on """
+
       def refresh(self) -> None:
       """
       Reload the cached task data from the task store
@@ -175,7 +181,7 @@ Backend implementors aren't required to implement their own ``TaskResult``, but 
       ...
 
 
-Attributes such as ``priority`` will reflect the values used to enqueue the task, as opposed to the defaults from the ``Task``. If no overridden values are provided, the value will mirror the ``Task``.
+Attributes such as ``priority`` and ``queue_name`` will reflect the values used to enqueue the task, as opposed to the defaults from the ``Task``. If no overridden values are provided, the value will mirror the ``Task``.
 
 A ``TaskResult`` will cache its values, relying on the user calling ``refresh`` to reload the values from the task store. An ``async`` version of ``refresh`` is automatically provided by ``TaskResult`` using ``asgiref.sync_to_async``.
 
@@ -299,11 +305,15 @@ Settings
    TASKS = {
       "default": {
          "BACKEND": "django.tasks.backends.ImmediateBackend",
+         "QUEUES": []
          "OPTIONS": {}
       }
    }
 
-``OPTIONS`` is passed as-is to the backend's constructor.
+
+``QUEUES`` contains a list of valid queue names for the backend. If a task is queued to a queue which doesn't exist, an exception is raised. If omitted or empty, any name is valid.
+
+``OPTIONS`` is passed as-is to the backend's constructor. ``QUEUES`` is additionally passed to the constructor as the ``queues`` keyword argument.
 
 Motivation
 ==========
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -303,7 +303,7 @@ The field of background tasks is vast, and attempting to implement everything su
 - Automated task retrying
 - A generic way of executing task runners. This will remain the repsonsibility of the underlying implementation, and the user to execute correctly.
 - Observability into task queues, including monitoring and reporting
-- Cron-based scheduling. For now, this can be achieved by triggering the same task once it finishes.
+- Cron-based scheduling
 
 Copyright
 =========
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -124,7 +124,7 @@ A ``Task`` is used as a handle to the running task, and contains useful informat
       status: TaskStatus
       """The status of the task"""
 
-      result: Any
+      result: Any | None
       """The return value from the task function"""
 
       queued_at: datetime
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -41,7 +41,7 @@ A backend will be a class which extends a Django-defined base class, and provide
    from django.tasks.backends.base import BaseTaskBackend
 
 
-   class MyBackend(BaseTaskbackend):
+   class MyBackend(BaseTaskBackend):
       task_class = Task
 
       def __init__(self, settings_dict: Dict):
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -90,7 +90,7 @@ A backend will be a class which extends a Django-defined base class, and provide
          """
          ...
 
-If a backend doesn't support a particular scheduling mode, it simply does not define the method. Convenience methods ``supports_enqueue`` and ``supports_defer`` will be implemented by ``BaseTaskBackend``. Similarly, ``BaseTaskBackend`` will provide stubs which expose ``enqueue`` and ``defer`` wrapped with ``asgiref.sync_to_async``.
+If a backend doesn't support a particular scheduling mode, it simply does not define the method. Convenience methods ``supports_enqueue`` and ``supports_defer`` will be implemented by ``BaseTaskBackend``. Similarly, ``BaseTaskBackend`` will provide ``a``-prefixed stubs for ``enqueue``, ``defer`` and ``get_task`` wrapped with ``asgiref.sync_to_async``.
 
 Django will ship with 2 implementations:
 
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -72,6 +72,18 @@ A backend will be a class which extends a Django-defined base class, and provide
          """
          ...
 
+      def get_task(self, task_id: str) -> BaseTask:
+         """
+         Retrieve a task by its id (if one exists)
+         """
+         ...
+
+      async def aget_task(self, task_id: str) -> BaseTask:
+         """
+         Retrieve a task by its id (if one exists)
+         """
+         ...
+
       def close(self) -> None:
          """
          Close any connections opened as part of the constructor
@@ -137,6 +149,7 @@ A ``Task`` is used as a handle to the running task, and contains useful informat
          """
          ...
 
+A ``Task`` is obtained either when scheduling a task function, or by calling ``get_task`` on the backend.
 
 A ``Task`` will cache its values, relying on the user calling ``refresh`` / ``arefresh`` to reload the values from the task store.
 
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -50,13 +50,13 @@ A backend will be a class which extends a Django-defined base class, and provide
 
       def enqueue(self, func: Callable, priority: int | None, args: List, kwargs: Dict) -> BaseTask:
          """
-         Queue up a task function to be executed
+         Queue up a task function (or coroutine) to be executed
          """
          ...
 
       def defer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
          """
-         Add a task to be completed at a specific time
+         Add a task function (or coroutine) to be completed at a specific time
          """
          ...
 
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -276,9 +276,9 @@ So that library maintainers can use this integration without concern as to wheth
 Reference Implementation
 ========================
 
-The reference implementation is currently being developed alongside this DEP process. This implementation will serve both as an "early-access" demo to get initial feedback and start using the interface, as the basis for the integration with Django core, but also as a backport for users of supported Django versions prior to this work being released.
+The reference implementation will be developed alongside this DEP process. This implementation will serve both as an "early-access" demo to get initial feedback and start using the interface, as the basis for the integration with Django core, but also as a backport for users of supported Django versions prior to this work being released.
 
-Once code is available, it will be referenced here.
+A more complete implementation picture can be found at https://github.com/RealOrangeOne/django-core-tasks, however it should not be considered final.
 
 Future iterations
 =================
--- a/draft/0000-background-workers.rst
+++ b/draft/0000-background-workers.rst
@@ -37,58 +37,36 @@ A backend will be a class which extends a Django-defined base class, and provide
    from datetime import datetime
    from typing import Callable, Dict, List
 
-   from django.tasks import BaseTask
+   from django.tasks import Task, TaskResult
    from django.tasks.backends.base import BaseTaskBackend
 
 
    class MyBackend(BaseTaskbackend):
+      task_class = Task
+
       def __init__(self, options: Dict):
          """
          Any connections which need to be setup can be done here
          """
          super().__init__(options)
 
-      def is_valid_task_function(self, func: Callable) -> bool:
+      @classmethod
+      def is_valid_task(cls, task: Task) -> bool:
          """
-         Determine whether the provided callable is valid as a task function.
-         """
-         ...
-
-      def enqueue(self, func: Callable, priority: int | None, args: List, kwargs: Dict) -> BaseTask:
-         """
-         Queue up a task function (or coroutine) to be executed
+         Determine whether the provided task is one which can be executed by the backend.
          """
          ...
 
-      def defer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
+      def enqueue(self, task: Task, *, args: List, kwargs: Dict, priority: int | None = None, run_after: datetime | None = None) -> TaskResult:
          """
-         Add a task function (or coroutine) to be completed at a specific (timezone-aware) time
-         """
-         ...
-
-      async def aenqueue(self, func: Callable, priority: int | None, args: List, kwargs: Dict) -> BaseTask:
-         """
-         Queue up a task function (or coroutine) to be executed
+         Queue up a task to be executed
          """
          ...
 
-      async def adefer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
-         """
-         Add a task function (or coroutine) to be completed at a specific (timezone-aware) time
-         """
-         ...
-
-      def get_task(self, task_id: str) -> BaseTask:
+      def get_task(self, task_id: str) -> TaskResult:
          """
          Retrieve a task by its id (if one exists).
-         If one doesn't, raises self.TaskDoesNotExist.
-         """
-         ...
-
-      async def aget_task(self, task_id: str) -> BaseTask:
-         """
-         Retrieve a task by its id (if one exists).
-         If one doesn't, raises self.TaskDoesNotExist.
+         If one doesn't, raises TaskDoesNotExist.
          """
          ...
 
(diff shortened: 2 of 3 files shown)


draft/0000-background-workers.rst
```diff
@@ -0,0 +1,299 @@
...
+
+.. code:: python
+
+   from django.contrib.tasks import task
+
+   # Later...
+   task = task.enqueue(do_a_thing)
+
+   # Clearer
+   thing_task = task.enqueue(do_a_thing)
```
RealOrangeOne, line 269: Naming things is hard - I'd love some alternative proposals for how to handle this naming. Even having the connection handler called `tasks` erks me out a little

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,299 @@
...
+         """
+         Close any connections opened as part of the constructor
+         """
+         ...
+
+If a backend doesn't support a particular scheduling mode, it simply does not define the method. Convenience methods ``supports_async`` and ``supports_defer`` will be implemented by ``BaseTaskBackend``.
+
+Django will ship with 2 implementations:
+
+ImmediateBackend
```
RealOrangeOne, line 85: Question: To follow Django's existing naming convention, would `DummyBackend` be a better name?

carltongibson:
Thanks so much for getting to this so quickly @RealOrangeOne! 🎁

I’ll have a proper read over the weekend but, one initial remark: I don’t think this should be in the `contrib` namespace. Either `django.tasks` or `django.core.tasks`. (I kind of prefer the former: `from django import tasks` would be 🥰)

prophile:
Hello! Thanks for putting this proposal up, IMO this would be great to have. Dropping some comments here for completeness (as we discussed in DMs), though I'm not sure this is the right place.

# Task/transaction race condition

In a backend which queues tasks onto an external queue (let's say, Redis), there's a race condition between a task runner starting the task and the transaction being committed. If I as a caller create a `PaymentRequest` entity and then enqueue a task to process it, the task runner might start processing the task before the transaction has committed, and so the task runner won't be able to see the `PaymentRequest` entity. The way to avoid the race condition is to use `transaction.on_commit` to enqueue the task only after the transaction has committed.

The problem is, beyond being a bit of a pain to do manually and being easily forgotten, it's not clear what is best to do in the `DatabaseBackend` here. For the `DatabaseBackend` the _opposite_ applies – because we're enqueueing the task inside the database, we want that to be transactional, and it's most unclear to me what the guaranteed semantics are of doing database writes in `transaction.on_commit`. For code which wants to portably support both backends, which I presume is the goal for Wagtail, this would presumably mean conditionally using or not using `transaction.on_commit` based on the backend, which is quite a pain, and doesn't necessarily generalise to new backends.

# Transactions in `ImmediateBackend`

There are some fairly important subtleties with transactions in `ImmediateBackend` which it's important to nail down in the non-autocommit case. Specifically - does the task run in the same transaction as the caller, and how?

## Case 1: Task runs in the same transaction as the caller

This is the 'obvious' implementation but comes with some major hazards.

First, if the task errors, this can leave the transaction in an inconsistent or aborted state, which is a fairly major footgun, not to mention a substantially different set of problems than would be experienced in production.

Second, in the case of asynchronous task runners, there is no guarantee that the caller will wait for the task to complete before closing the transaction, which would leave the task in a very peculiar state.

## Case 2: Task runs in the same transaction as the caller, but with a savepoint

This is a bit more robust, but only limits rather than fixes the first problem, and does nothing for the second. Releasing a savepoint is also not the same thing as committing a transaction, and so this can't replicate on-commit behaviours either from the database or Django. The former can pop up errors on commit which won't show up here, and the latter matters for `on_commit` calls because these will then be run when the _caller_ commits, not when the task completes.

## Case 3: Task runs in a separate transaction

This provides the most robust semantics, and best mirrors the production use, but also opens up two pretty big cans of worms:

1. How is this even implemented? A separate connection to the database? Are we talking threads? Does that work for both synchronous and asynchronous code? What about the `transaction.on_commit` problem?
2. What does this mean for tests? Tests are likely to be one of if not _the_ major use of the `ImmediateBackend`, but for fast tests we definitely want to be rolling back transactions after each test rather than doing the `TRUNCATE`s manually to reset each table. How would or could that interact with the separate transactions for the task? Can asynchronous, or even synchronous, tasks dangle after a test has finished?

# Serialization in `DatabaseBackend`

I presume that the `args` and `kwargs` are serialized with Pickle in the `DatabaseBackend`? Some considerations:

1. This is probably worth documenting explicitly, because it might be something of a surprise if switching to a different backend.
2. Will the `ImmediateBackend` replicate the same serialization? If not, this could be a source of surprises when switching between backends. I'm also not sure of the implications of passing ORM objects as arguments to tasks and the difference between passing them with and without intermediate serialization.
3. Moving a class to a different module, or a few other innocuous changes, will cause task dequeuing to suddenly break after a release for tasks enqueued beforehand. This is potentially quite surprising. Perhaps there's something to be learned from the serialization infrastructure for sessions, which have quite similar problems?
4. If this used Pickle protocol version -1 (i.e. the latest) then there's a race condition around releases if a new version were released – producers would immediately start enqueuing tasks with the new version, but consumers would still be running the old version. Should the version be pinned? This is maybe not fundamentally different from the problem of rolling out new versions of classes which are being pickled and passed around in general, but it's also potentially much more surprising to developers if it happens "behind their backs". For instance, if you upgraded Python version in production, you might unexpectedly not be able to roll back because all the enqueued tasks are unexpectedly not backwards-compatible.

# Standard priority ordering

Your priority ordering is an `int` but there are differing standards for what the ordering should be. There's quite differing prior art here:

Lower is more important: UNIX (`nice` levels), Celery (Redis backend)

Higher is more important: Windows thread scheduling priorities, AMQP (including RabbitMQ), Celery (RabbitMQ backend)

It seems like it would be a good idea to standardise on one of these because otherwise there's no way to portably do prioritisation, and portable prioritisation is presumably why that argument is there in the first place.

For backends which can't support priority, which includes potentially fairly major targets like SQS and Google Cloud Pub/Sub, what's the behaviour if you pass a priority argument? Is it silently ignored, and is that safe?

# Dedicated capacity

This configuration semantically puts all tasks into one (priority) queue. In previous projects I (and presumably others) have relied on being able to provision dedicated capacity with things split into multiple named queues, even if there's a default. Having a way to (1) specify this portably and (2) implement it with the `DatabaseBackend` would be fairly important for me. YMMV.

# Reliability semantics and retrying

For automatic task retrying, whenever I've worked with queue infrastructure which doesn't have reliability, I've eventually found it necessary to emulate it. If this has been others' experience (and I'm not a terribly interesting developer so if I've experienced this I suspect a lot of people have), it would be good to take the opportunity here?

It also affects the design of the interface, so IMO it's at least worth spelling out how this would work even if the implementations are left for later.

When a task fails, it's necessary to have some way of signalling whether or not it should be retried. If my email sending task fails because the SMTP server is down, I probably want to retry it; if it fails because the email address is invalid, or because I can't actually decode the task data due to a Pickle upgrade, I probably don't want to retry it. The specification here doesn't have a way to portably signal that though, and `TaskStatus` has only `Failed` in the portable subset, where it may be useful to differentiate retryable and non-retryable failures.

# Task status signalling in `DatabaseBackend`

Given portable transactions, in the `DatabaseBackend`, what are your expectations around when and how the task runner locks the task instances? If the transaction for a task which is `Running` fails, how do you update its status afterwards from `Running` to `Failed` without a race condition?

# Entry cleanup in `DatabaseBackend`

Task records are presumably not left around forever otherwise the amount of data used grows without bound. What's the strategy to clean up tasks records for tasks that have completed, and particularly what's the strategy to do this without a race condition?

# Final thoughts

I'm asking these questions not because I object to the proposal but exactly because I'd like to see this proposal succeed. Give me a shout if there's anything I can clarify or help with here.

draft/0000-background-workers.rst
prophile: IMO I'd expect a `DummyBackend` to throw away tasks (potentially with logging) in the same way as it would in the cache backend–in that sense `ImmediateBackend` isn't a dummy, it's actually doing work.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,299 @@
...
+=================
+
+The field of background tasks is vast, and attempting to implement everything supported by existing tools in the first iteration is futile. The following functionality has been considered, and deemed explicitly out of scope of the first pass, but still worthy of future development:
+
+- Completion hooks, to run subsequent tasks automatically
+- Bulk queueing
+- Automated task retrying
+- A generic way of executing task runners. This will remain the repsonsibility of the underlying implementation, and the user to execute correctly.
+- Observability into task queues, including monitoring and reporting
+- Cron-based scheduling. For now, this can be achieved by triggering the same task once it finishes.
```
BugliL, line 293: 👍🏼

jacobian:
Thank you **so much** for putting this together, @RealOrangeOne! This has long been on my personal Django wishlist, so I'm absolutely stoked to see it moving forward.

I've just had a quick read, so don't have super-detailed comments, but I do have a couple of big-picture questions that I don't see adressed, and that I think need to have some answers:

1. How "far" do you picture this framework going? That is: Celery (to pick the most full-featured Python queue) has just a huge ton of features. There's probably a point at which if you need some of those things, you should just be using Celery. I think having some idea of where the "line" is going to be would be helpful. 

   Personally, I think whatever's built in to Django should aim to be super-simple, with minimal configuration, and a minimally-useful subset. Beyond that, I think we should be pushing people to more full-featured external libraries. That is, for me, a built-in task queue needs to handle two major use-cases:
   
   - Facilitate rapid early development - when you're starting a new app, having to set up a more complex task queue can be a barrier. Having something built-in means you can quickly do background email sending (or whatever) without too much faff. This is the "... with deadlines" part of the Django tagline, right?
   - Give third-party apps a common baseline, so they can expect a task queue without needing to commit to a specific dependency. This is the Wagtail use-case, I think, right?
   
   There may be other use-cases -- but the point is i think we need to enumerate them to avoid feeling like we need to build something feature-competitive with Celery or whatever. 

2. Related: what is the "transition path" for moving from the built-in task queue to something else? Some amount of refactoring is unavoidable and fine, but I think the DEP should address, at least a bit, how we expect users to transition if they outgrow what the built-in queue offers.

RealOrangeOne:
> How "far" do you picture this framework going?

No where near Celery. The vast majority of this proposal is about writing the glue which sits between Django and tools like Celery, so developers and library maintainers have a single interface to using queues, without needing to worry about the implementation details. 

The biggest risk of scope-creep for this is definitely the `DatabaseBackend`, as it actually tries to implement a task queue (which as @prophile outline above, is quite difficult). I agree that I see `DatabaseBackend` as being perfect for anything which needs a sensible and performant job queue. If people already know they need Celery, it's absolutely not the backend for them, and I don't think it should try to. With that said, I'd like to try and avoid intentionally nerfing the `DatabaseBackend` - just because it's optimised for smaller deployments, doesn't mean it shouldn't be capable of handling a very large number of tasks concurrently, and operate in a sensible and expected way. I suspect the target audience for `DatabaseBackend` and a SQL-based queue are fairly well aligned with those who may choose something like PostgreSQL FTS over something like ElasticSearch. Sure, ElasticSearch is probably better for those 10% of users who really need it, but doesn't mean the other 90% won't be perfectly happy with postgres, and probably wouldn't benefit from ElasticSearch anyway.

I think the `DatabaseBackend` both helps people get moving quickly (and scale up fairly far without needing to worry _too_ much). Once someone needs to move, there's little application code which needs changing to move over to something like Celery should its wealth of features be needed.

> what is the "transition path" for moving from the built-in task queue to something else

I think there are 2 things which would need to transition:

Code wise, I'd like there to be as close to 0 changes needed to swap out the backend (up to a certain point). If implementation details are encoded into how tasks are queued (eg the need for `transaction.on_commit` for non-DB queues), then that may need changing. Otherwise, I see a world where the only changes are those in `settings.py`, much like we have now with `django.core.cache`.

As for the queue storage itself, I think that's a little more complex. Exactly how to say move a bunch of tasks from RQ to Celery and RabbitMQ is probably quite a difficult story, which doesn't have an obvious answer. However, I don't think that needs to be on Django to solve, since it's only providing the glue between the runner and Django, as opposed to a library for other task runners to integrate with (Other than implementing this interface, I'd like there to be no changes Celery would need to make to support this). The easiest transition path might be running both queues in parallel for a while until the first queue is empty, and then stopping the runners.

orf:
I love this, thank you @RealOrangeOne. A couple of thoughts:

Passing a plain callable to `enqueue` and `defer` is a nice interface but it has some downsides. We can only *really* enqueue/delay globally importable functions: no lambdas, no local functions, etc. Handling this at the callsite or the receiver both have downsides, and it’s generally confusing to new users. What do we do when a user tries to enqueue the callable `some_model_instance.send_email`?

Should we add some concept of ownership/queue/routing to the task model itself? Given a task ID how do I know what queue it was sent to, or how would I send a task to a specific queue? Defining lots of duplicate backends (1 per queue) seems redundant and not very performant.

An execution of a task is a separate thing to the task itself - a task might have 0 or more executions (i.e retries, timeouts, etc). The current model seems to mix both of these together - is it worth making a distinction?

Can we/should we tie enqueuing into “transaction.on_commit” by default? Seems like a big footgun people will run into almost immediately.

I personally hate `enqueue` and `defer` - they are pretty widely used, but `run_later` and `run_at` seem more approachable.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,312 @@
...
```
smark-1, line 306: How would the initial Cron tasks be scheduled? Or is this only meant for tasks that start based on an event?

Wagtail recommends a few management commands to be run like: `publish_scheduled` every hour and `update_index` once a week.
I assume that the Cron-based scheduling would be able to run these types of tasks.

jacobian:
> I suspect the target audience for DatabaseBackend and a SQL-based queue are fairly well aligned with those who may choose something like PostgreSQL FTS over something like ElasticSearch. Sure, ElasticSearch is probably better for those 10% of users who really need it, but doesn't mean the other 90% won't be perfectly happy with postgres, and probably wouldn't benefit from ElasticSearch anyway.

This is _fantastic_ framing, and something that should probably end up in the eventually docs for this. Super-well-said!

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,312 @@
...
+========
+
+Django doesn't have a first-party solution for long-running tasks, however the ecosystem is filled with incredibly popular frameworks, all of which interact with Django in slightly different ways. Other frameworks such as Laravel have background workers built-in, allowing them to push tasks into the background to be processed at a later date, without requiring the end user to wait for them to occur.
+
+Library maintainers must implement support for any possible task backend separately, should they wish to offload functionality to the background. This includes smaller libraries, but also larger meta-frameworks with their own package ecosystem such as `Wagtail <https://wagtail.org>`_.
+
+Specification
+=============
+
+The proposed implementation will be in the form of an application wide "task backend" interface. This backend will be what connects Django to the task runners with a single pattern. The task backend will provide an interface for either third-party libraries, or application developers to specify how tasks should be created and pushed into the background
```
lb-, line 28:
```suggestion
The proposed implementation will be in the form of an application wide "task backend" interface. This backend will be what connects Django to the task runners with a single pattern. The task backend will provide an interface for either third-party libraries, or application developers to specify how tasks should be created and pushed into the background.
```

PeterJCLaw (commented):
Thanks for putting this together :clap:. It's really cool to see something like this coming to Django.

I've added some comments & questions inline, though I also have some thoughts towards an alternative API which alleviates some of the things I and others have raised. I suspect this is into the realm of _bigger_ thoughts so I'll save that for another venue. (The short version is that it's somewhere between DLQ's API and Django's ORM). I'll DM you somewhere to chat.

For context: my experience with queueing systems is almost entirely with somewhat home-grown and less well known one (https://github.com/thread/django-lightweight-queue, DLQ), which I suspect has fewer features than the libraries already mentioned. As a consequence I've had some hands-on experience extending such a system. Many of the comments I've added are built on learnings from that.

Just a thought -- would it be worth inviting the maintainers of Celery/RQ/etc. to have a look at this PR? Especially if the goal is that transitioning from Django's internal system to one of those is relatively easy I suspect they could have a lot of wisdom to share.

draft/0000-background-workers.rst
smark-1: Using the ready method might cause some issues since the ready method gets run on each server so it can cause the scheduled tasks to run multiple times. Adding a get_or_create method can fix this.

Using the apps ready method for Cron based scheduling sounds good for now. For future iterations it would be nice if each Cron Task can inherit from a `CronJob` base class similar to how https://github.com/Tivix/django-cron does it. Maybe even autodetect a tasks.py file and auto register these Cron tasks.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,312 @@
...
+
+Backends
+--------
+
+A backend will be a class which extends a Django-defined base class, and provides the common interface between Django and the underlying task runner.
+
+.. code:: python
```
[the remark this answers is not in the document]
RealOrangeOne: The type hints are much more about making the reference clearer than actually being the canonical types.

I too don't know what Django's policy is, but personally my intention is to try and make the API type-safe, and the types themselves can either be in the codebase or external (probably the latter). That way adopting types is simple, without necessarily needing to adopt a typechecker as part of this DEP.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,312 @@
...
+         """
+         Add a task function (or coroutine) to be completed at a specific time
+         """
+         ...
+
+      def get_task(self, task_id: str) -> BaseTask:
+         """
+         Retrieve a task by its id (if one exists)
+         """
+         ...
```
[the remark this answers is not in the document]
RealOrangeOne: It'll raise a generic exception (ie not backend dependent), yes. I agree that should be made more explicit.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,312 @@
...
+
+.. code:: python
+
+   from django.utils import timezone
+   from datetime import timedelta
+   from django.tasks import defer
+
+   task = defer(do_a_task, when=timezone.now() + timedelta(minutes=5))
+
+When scheduling a task, it may not be **exactly** that time a task is executed, however it should be accurate to within a few seconds. This will depend on the current state of the queue and task runners, and is out of the control of Django.
```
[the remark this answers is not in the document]
RealOrangeOne: I think this may be a layover from a previous verison of this DEP, where it was more important to define this. It's absolutely backend dependent, but I still wanted to flag both that there's little Django can do, and that there's a risk anyway (a risk that is probably the case anyway).

draft/0000-background-workers.rst
RealOrangeOne: This is a great point, there's a bootstrapping issue for that cron implementation. I think I'll remove the hint, as in reality it's much more complex than that.

I suspect the current pattern of an external scheduler (ie `crond`) running management commands is probably the way to go, at least for now.

draft/0000-background-workers.rst
RealOrangeOne: The original version used `background`, but I changed it for exactly that reason. 

"Runner", "Worker" and "Queue" are all fairly specific components in the task execution pipeline, and that's not what this really is. I _guess_ a task backend could be considered adding a task to a "Queue", but when developing a backend, having 2 very different things called "queues" is quite annoying.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,312 @@
...
+
+      def close(self) -> None:
+         """
+         Close any connections opened as part of the constructor
+         """
+         ...
+
+If a backend doesn't support a particular scheduling mode, it simply does not define the method. Convenience methods ``supports_enqueue`` and ``supports_defer`` will be implemented by ``BaseTaskBackend``. Similarly, ``BaseTaskBackend`` will provide stubs which expose ``enqueue`` and ``defer`` wrapped with ``asgiref.sync_to_async``.
+
+Django will ship with 2 implementations:
```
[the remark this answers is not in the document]
RealOrangeOne: I really like this idea. `DatabaseBackend` does work, but I agree for test performance alone it'd be nice to have something purely in-memory (akin to `InMemoryCache`).

By default, `get_task` would be the main interface to the internal task queue, but there's on reason just the in-memory backend couldn't also expose a `.queue` which contains everything (sorted by priority, probably).

draft/0000-background-workers.rst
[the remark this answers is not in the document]
RealOrangeOne: Implementation wise, they're very similar, sure. Definition wise however they're quite different. `when` is just a "do not execute before X" field in the queue. cron however not only requires the list of tasks to be defined statically (to avoid a bootstrapping issue), but also some kind of lock to confirm 2 workers don't try to run it at once.

draft/0000-background-workers.rst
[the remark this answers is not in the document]
danpalmer: Should the test performance aspect here be delegated to the database backend, assuming SQLite in-memory if people are performance conscious?

draft/0000-background-workers.rst
rik: Taking inspiration from the [Rails ActiveJob API](https://guides.rubyonrails.org/active_job_basics.html), I quite like this API:

```py
from django.tasks import perform_later

perform_later(send_email)
perform_later(send_email, wait_until=timezone.now() + timedelta(minutes=5))
```

Only exposing one function (and its async counterpart) avoids having to come up with different names.

Leaving `run_later`, `call_later`, `execute_later` as alternatives.

draft/0000-background-workers.rst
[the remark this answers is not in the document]
RealOrangeOne: In-memory SQLite would partially solve the performance issue (purely in-memory is still going to be faster without the DB overhead).

However, if `DatabaseBackend` implemented a way of getting the entire queue, we'd need to think carefully about the expectations of other backends supporting it, which isn't the case with a purely in-memory backend, which quite clearly is only useful during tests.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,329 @@
...
+      """The return value from the task function"""
+
+      queued_at: datetime
+      """When the task was added to the queue"""
+
+      completed_at: datetime | None
+      """When the task was completed"""
+
+      raw: Any | None
+      """Return the underlying runner's task handle"""
```
ryanhiebert, line 135: Should we standardize this? `raw` is a frustratingly generic name, and backends can return subclasses that include whatever other properties they want. Accessing it to do anything useful will always be a backend-specific operation.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,329 @@
...
+      """A unique identifier for the task"""
+
+      status: TaskStatus
+      """The status of the task"""
+
+      result: Any
+      """The return value from the task function"""
+
+      queued_at: datetime
+      """When the task was added to the queue"""
```
ryanhiebert, line 129: Are any of these properties optional to implement? For example, I could imagine wanting to write a minimal backend that doesn't have statuses or results, because they don't fit in neatly with the broker I'm using for queuing.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,329 @@
...
+         """
+         Reload the cached task data from the task store
+         """
+         ...
+
+A ``Task`` is obtained either when scheduling a task function, or by calling ``get_task`` on the backend. If called with a ``task_id`` which doesn't exist, a ``TaskDoesNotExist`` exception is raised.
+
+A ``Task`` will cache its values, relying on the user calling ``refresh`` / ``arefresh`` to reload the values from the task store.
+
+To enable a ``Task`` to be backend-agnostic, statuses must include a set of known values. Additional fields may be added if the backend supports them, but these attributes must be supported:
```
ryanhiebert, line 171: I'm finding this line confusing. Are you saying that these well-defined status states must be supported (if statuses are supported at all)? Or are you saying that all of the above attributes defined in the task definition must be supported (so status, result, etc are non-optional and backends _must_ implement them to be compliant?

draft/0000-background-workers.rst
ryanhiebert, line 1: We should declare the expected acknowledgement semantics.

While I very heavily use `acks_late=True` in combination with reliable message broker like RabbitMQ, I think that if we want to keep a simpler interface we should define that they are acknowledged before work on the task begins. Thus, we should document that backends are expected to run tasks **at most once** in order to be compatible with this interface. Senders of tasks (including library authors) that perform such dangerous actions as sending out emails need to be confident that they aren't going to be sent out multiple times.

At least once execution is also a very helpful semantic for a broad variety of tasks, and I hope that we can standardize an interface for that in the future. What I think is **unwise** would be to leave this semantic distinction undefined in the specification.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,329 @@
...
+
+:New: The task has been created, but hasn't started running yet
+:Running: The task is currently running
+:Failed: The task failed
+:Complete: The task is complete, and the result is accessible
```
ryanhiebert, line 176: Are additional states not listed here acceptable for backend implementations to return? For example, Celery runners will prefetch jobs off the queue for efficiency, would it be acceptable for a similar implementation to add a `RESERVED` status? How should callers expect to handle that?

draft/0000-background-workers.rst
ryanhiebert, line 1: Celery may not be able to be a straightforward backend as-is. And I'm not sure that's a problem we want to fix. Celery, afaik, requires tasks to be registered in the worker. But the more I think about the interface, the more I like that this _doesn't_ require the callables to be registered somewhere, and I hope we can keep that.

ryanhiebert, line 1: Database backends expect to be connecting to one backing service. Cache backends have the same assumption. However, if I want to make the most robust implementation of this interface, I'm very likely to need to connect to multiple backing services, most often RabbitMQ for queueing and Redis for results and statuses.

Noting this dissimilarity had made me wonder if we should have a conception of a queue broker or a result backend or a status backend. And I think not. But we should make that trade-off explicitly. W want to encourage the very best implementations of this interface, but we don't want to be defining that much of the complexity that these systems often grow in this interface.

ryanhiebert, line 1: It may be worth pointing out in the document that backends generally are *not* the correct mechanism for creating multiple queues. Like with databases, we support multiple backends, but we want library authors to be able to delay tasks without essentially requiring them to have configuration to figure out which backend to use.

We aren't currently defining semantics for how backends will assign queues, but successful projects typically need multiple queues long before they need multiple databases, and often even longer before they will need to have multiple different queuing *systems* for their background tasks.

With this interface as it is currently defined, some backends should be able to implement queues based on the callable's path. I tend to think that this means of queuing can get most projects quite far, so I'm not sure how much an explicit interface for queuing is necessary here. It would be sufficient for the use-case I've most commonly encountered, where long-running callables end up blocking what would otherwise be short-running callables from even getting to a runner.

ryanhiebert (commented):
Thank you so much for bringing this forward. I fully support it. I've spent the past two weeks battling myself on what things I think are critical features for this interface.

ryanhiebert:
> > Should we add some concept of ownership/queue/routing to the task model itself
> 
> Yes, absolutely we should. A `Task` should probably know all the attributes which were passed to the `enqueue` call which triggered it.

I think that queuing under the backend interface described would rightly be an implementation detail of the backend, and can actually be reasonably capable. The backend, for example, can separate callables into queues based on the import path. That would get a lot of the projects I've worked on a good deal of mileage, to support the common case of separating long-running callables from short-lived callables.

If ownership/routing is anything like exchanges and routing keys in Celery / AMQP, I think those are advanced use-cases that we don't want to get into.

draft/0000-background-workers.rst
orf: However this means that moving a task function breaks all in-flight tasks, plus all the lovely security issues that come from a remote source giving us arbitrary functions, by path, to call.

Registration is definitely the way, unfortunately.

orf: Just to elaborate on this, what if a worker receives a malicious message like so:

```json
{“function”: “subprocess.check_call”, “args”: [“rm”, “-rf”, “/“]}
``` 

There has to be a form of task registration, so that there is an allowlist of specific functions to run when a task is received. This also helps decouple the functions import path from the actual message, which is a good thing ™️

ryanhiebert: I don't think that we should attempt to make this interface resilient to malicious messages any more than Django generally make things resilient to malicious database queries. Sending messages is dangerous, and it is intended to do dangerous things, and you need to make sure you trust the code that can do that.

It's true that you can't change the location of a callable all at once without the possible need for downtime, but that's pretty easy to deal with by creating a wrapper function in one place that calls the other.

orf: > Sending messages is dangerous, and it is intended to do dangerous things, and you need to make sure you trust the code that can do that.

There is a _vast_ difference between this and "anyone who can add messages into the queue now has full remote-code execution capabilities, no questions asked". Invoking _completely arbitrary_, user-supplied functions with user-supplied arguments from outside sources has always ended rather poorly.

There are also unresolved issues around how to handle `task.enqueue(some_model_instance.send_email)`. Some things can't, and shouldn't, be done with this kind of pass-an-arbitrary-callable interface.

orf: IMO a decorator that does task registration solves a bunch of these issues without impacting usability. We can use it to handle the _very_ common case of "i want to pass a model instance to a task" where the PK is sent over the wire, but the task decorator fetches the instance it before calling the function. This would neatly support model methods that can also be tasks 🤷.

```python
# look mum, no RCE!
class SomeModel:
    @task.task
    def send_email(self):
      ...
```

_This_ is the expected dangers of a task queue - you can send arbitrary IDs to tasks. A `do_dangerous_thing(model: Model)` task signature could be called with arbitrary models. This differs significantly from blindly allowing and executing `psql -c 'TRUNCATE TABLE important_data'` because a message told us to.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,329 @@
...
+         """
+         ...
+
+      def close(self) -> None:
+         """
+         Close any connections opened as part of the constructor
+         """
+         ...
+
+If a backend doesn't support a particular scheduling mode, it simply does not define the method. Convenience methods ``supports_enqueue`` and ``supports_defer`` will be implemented by ``BaseTaskBackend``. Similarly, ``BaseTaskBackend`` will provide ``a``-prefixed stubs for ``enqueue``, ``defer`` and ``get_task`` wrapped with ``asgiref.sync_to_async``.
```
ryanhiebert, line 93: Would it be reasonable to have the reverse auto-implementations as well, so that backends can be async first and get this default implementation? It's a bit trickier because the implementations will need to figure out a good way to break the recursion, but if this is for convenience rather than backward compatibility the convenience factor seems to apply in both directions.

draft/0000-background-workers.rst
orf: > This is roughly analogous to giving someone direct access to the database, although since it doesn't (necessarily) require access to the database it's not an identical concern, and a security-in-depth approach may well find it beneficial to add restrictions to require task registration. I think it can be a reasonable concession to simplicity to recognize that parallel and make this trade-off toward a simpler interface option.

Sure, there are definitely trade-offs to make here, but I strongly disagree that it's analgous to giving access to the database. In a simple, single-service world where exactly the same Django application is both sending and processing messages then I can understand this position. However there are more things to consider here.

Messaging queues are often used to connect two separate systems together, across different domains and differing responsibilities. Even if those systems share a common code-base, it seems like shaky ground to design something that allows simple lateral movement across services like this.

Secondly, brokers have differing security models. Take a hypothetical example of using SQS as a broker - anyone or any service with the `sqs:SendMessage` IAM permission now has complete access to any Django system processing those messages. That would be highly unexpected, as the ability to _send_ a message (but not receive it) is something that is fairly liberally given between services, areas of responsibility or even AWS accounts.

However, all this might be redundant. We seem to be assuming that there is a binary choice here: registration _or_ flexibility. Why can't we have both? We don't _need_ registration, we just need an allowlist, so why don't we use a decorator for this? 

```python
# Inside Django
def run_task(message):
    callable = import_string(message.function_location)
    if not isinstance(callable, Task):
        raise SomethingSketchyIsGoingOn()
    callable(message.args)

# Anywhere
@task.task
def my_task(...): ...
```

This avoids the need to dynamically register all tasks, and pushes the dodgy aspect from "anyone sending messages can run any code they want" to "anyone sending messages can import a module". This seems like a much nicer security boundary, and we can be smart about only importing modules from `INSTALLED_APPS`.

ryanhiebert: > > This is roughly analogous to giving someone direct access to the database, [...]
> 
> [...] I strongly disagree that it's analogous to giving access to the database. [...]

You're totally right, and I don't want to give anyone the impression that the concerns you brought up aren't valid. What I was pointing out is that other types of backing stores aren't something you can be lax with security over, and the same is true for messaging queues. I know you agree with this. Trying to minimize the surface area of vulnerabilities to any particular store is still wise, and arbitrary callables referenced there can be a lot of room for significant fallout if a broker is compromised.

> In a simple, single-service world where exactly the same Django application is both sending and processing messages then I can understand this position. However there are more things to consider here.
> 
> Messaging queues are often used to connect two separate systems together, [...]

I really appreciate you saying this, that you see where I'm coming from, and you've hit the nail on the head about where our perspectives differ. In my opinion, the single-service world is what this API is and should be aiming for. A messaging system that is glue between disparate services is a categorically different type of system, and I do not think that this API should aim to be a worthwhile abstraction in that type of system.

This is a simple battery for running background jobs, and it shouldn't be used to implement an event-driven architecture.

> However, all this might be redundant. We seem to be assuming that there is a binary choice here: registration _or_ flexibility. Why can't we have both? We don't _really_ need registration, we just need an allowlist, so why don't we use a decorator for this?

I love it. You have found a great path that provides the allowlisting feature you're advocating without requiring registration which is what I'm really opposing. I'm on board. Marking a task as a task isn't burdensome, and having the decorator makes for an obvious extension point for future additions to this API.

It doesn't solve the Celery-compatibility challenge by itself, but combined with backend configuration to explicitly find these callables / tasks / we-need-a-glossary, I think this will be a great simple API with some reasonable security controls.

PeterJCLaw: Thanks both for exploring this aspect -- there's some good discussion here.

Personally I favour the registration approach. For me the security argument is pretty strong, though I do have other reasons to prefer it (two key ones are: clarifying that e.g: lambdas can't be used, and being able to attach data to the registration which then flows into the operation of the task -- an example might be a timeout or retry config).

I do appreciate the question around whether the failure mode here requires direct access to the underlying broker. I'd draw the comparison to how Django protects developers from e.g: SQL injection by ensuring that arguments are quoted when being passed to the database. In the instruction/data split, to me the task being run is "instruction" and the arguments to it are _maybe_ (hopefully) reasonable to consider as "data".

Even in a world where Django is the only expected producer & consumer of messages, as a developer I would prefer to be protected from foot-guns where possible.

With regards to the thoughts around registration vs allowlists -- I'd suggest picking a single interface for simplicity. IMO an allowlist can be easily enough simulated by users given a registration style API, either by implementing something similar to what's discussed above or by creating trivial wrappers. Perhaps having an example of doing this included in the docs would suffice?

draft/0000-background-workers.rst
RealOrangeOne: Defining one to call the other is going to lead to odd recursion issues. I think defining the sync API with an `async` stub is ok for now, and matches how the rest of Django works. As time goes on, we might want to remove the stub entirely and enforce the user to do it. For now, I think the `async` stub is the way to go.

draft/0000-background-workers.rst
RealOrangeOne: "at most once" semantics are definitely the easiest to reason about, although in some cases people may want the latter. I'm not exactly how we encode that in the current API. But I think I agree that we can assume "at most once" for all backends, and if we want it to be configurable over time, it'll probably end up being another argument to `enqueue` (or perhaps something defined at config time if it's not sensibly / easily configured per-task).

RealOrangeOne: I'd intentionally not added registration, as IMO calling `enqueue` directly is incredibly clear what's going on and "There should be one-- and preferably only one --obvious way to do it". RCE issues from the queue storage are a concern, however the same is true for anything storing arguments using `pickle`. Signatures would be a way to solve both, but that adds more constraints for backing workers. 

If Celery requires task functions to be marked, then it's unlikely they're the only library, and otherwise would make Celery impossible to integrate with. The decorator can defer how to "mark" a task function to the backend, whether that's by marking it with "this is an allowed method", or something more complex like Celery probably does. The decorator would be thin, and likely noop unless there's anything more complex. The idea being that the backend doesn't (and thus shouldn't) keep track of the registered tasks, as some modules may not have been loaded yet.

During enqueue, we check the callable has a hidden attribute which defines it can be run in a queue. If it doesn't, reject it immediately (and don't queue it). We can then provide a method for `is_valid_task_function`, which allows worker processes to confirm the callable is a valid task. The function would still need to be imported, so it wouldn't be "free", but still pretty cheap, and fairly reliably remove any RCE potential.

We _could_ make this opt-in functionality, or even opt-out, but personally I'd prefer Django be forcing good practice on a user, rather than require they opt-in to security, or allowing them to create a foot-gun and potential CVE for Django later on.

RealOrangeOne: I don't think this is necessarily a limitation in the current pattern. If you wanted to store results in `redis`, then so long as you configured your worker to write there (currently out of scope of this proposal), then you could easily override `get_task` to read from there. The configuration options are passed directly to the constructor, so you could include credentials for both RabbitMQ and Redis.

Sure, in the initial pass, I'm not really targetting this more complex _split-brain_ approach, but I don't think the current API makes it completely impossible.

RealOrangeOne: So far, all the times I've needed multiple queues has been to facade priorities into a system which doesn't quite support it in the way I need. What other uses of multiple queues do you see? Are you thinking the runners might be scaled slightly differently (akin to microservices), or is this more for load balancing.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,329 @@
...
+
+   class MyBackendTask(BaseTask):
+      id: str
+      """A unique identifier for the task"""
+
+      status: TaskStatus
+      """The status of the task"""
+
+      result: Any
+      """The return value from the task function"""
```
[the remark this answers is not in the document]
RealOrangeOne: I'd originally intended for it to just be `None`, but I agree I think it should probably raise an error, so it can be distinguished from "there's no result _yet_" and "the result is None". Although, arguably you can distinguish that with the status. 

I'm not sure what the most sensible way forward here is.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,329 @@
...
+         """
+         Queue up a task function to be executed
+         """
+         ...
+
+      def defer(self, func: Callable, priority: int | None, when: datetime, args: List, kwargs: Dict) -> BaseTask:
+         """
+         Add a task to be completed at a specific time
+         """
+         ...
```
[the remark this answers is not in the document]
RealOrangeOne: Damn timezones!

I think enforcing aware is probably the easiest way to go. Enforcing naive is going to lead to a lot of confusing bugs, and enforcing UTC doesn't feel like it really makes enough difference to just enforcing fully aware.

draft/0000-background-workers.rst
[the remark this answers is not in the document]
ryanhiebert: I think raising makes more sense. If they ask for the result, they probably want the result rather than a placeholder. If they want to avoid an exception, they should be checking the status to see whether the result is ready for them.

draft/0000-background-workers.rst
RealOrangeOne: It looks like [Laravel](https://laravel.com/docs/10.x/queues#delayed-dispatching) uses a similarly unified dispatch API, so unifying the call API could be nice. There's still a discovery API for whether scheduling is supported, so we can just reuse.

`perform_later` doesn't quite solve what to call the overarching connection manager, though.

draft/0000-background-workers.rst
ryanhiebert: It's possible that I just haven't given it a fair enough shake, but I've been afraid of relying on priorities for that purpose. For example, I might have a capacity of 8 active tasks over two runners, which generally has a low volume of long-running tasks with occasional flurries. If there are no high-priority tasks at some moment, all 8 could end up running the long-running tasks, blocking subsequently queued up high-priority tasks. If I have a queue that is reserved for _only_ running these high-priority tasks, then I can be sure that they aren't going to be entirely blocked from running by lower-priority tasks.

In RabbitMQ, available priorities have to be declared up front, because they are implemented as independent queues. With that knowledge, I think that a mechanism to specify that the runner should only run tasks with a priority X or above might fit my needs, and could be a mechanism for avoiding more explicit queuing. Since we're not specifying anything about the runner right now, that might make it reasonable to skip specifying alternative queues, at least for now.

draft/0000-background-workers.rst
RealOrangeOne: What standardization did you haven in mind? My intention behind this was for it to function as an ejection interface, such that a developer can easily access the underlying object if they wish. Less useful for application developers, but I still wanted it as part of the public API.

If there's something standard which is going to be commonly needed, it should be exposed in a standard API on the `Task`. Otherwise, people can use `raw`, but all bets are off.

draft/0000-background-workers.rst
RealOrangeOne: I've changed my mind on this point, mostly from a typing and casting safety perspective.

draft/0000-background-workers.rst
ryanhiebert: Sorry, I edited my comment a moment ago, I was suggesting we not bother including it _because_ it's an implementation detail. No big deal if you disagree.

draft/0000-background-workers.rst
ryanhiebert: > If Celery requires task functions to be marked, then it's unlikely they're the only library

I'm certain it is not the only one. [Dramatiq](https://dramatiq.io/) also requires actors (analog to Celery tasks) to be registered.

> personally I'd prefer Django be forcing good practice on a user, rather than require they opt-in to security, or allowing them to create a foot-gun and potential CVE for Django later on.

I agree.

ryanhiebert: What I've suggested makes conceptual sense, but for example I don't see that Celery supports running workers that only listen to certain priority levels. Restricting some runners to avoid low-priority tasks feels like a really nice interface and feels like it could make a simpler overall API with reasonable effectiveness, but I'm not sure there's any prior art to rely on to get us there.

manfre: > We could make this opt-in functionality, or even opt-out, but personally I'd prefer Django be forcing good practice on a user, rather than require they opt-in to security, or allowing them to create a foot-gun and potential CVE for Django later on.

Security needs to be on by default. We already receive emails to security@ that boil down to "When I ignore the security warning in the docs, there is a security gap". Any opt-out would need to be documented with very clear "Security warning" to scare away some who don't actually need to use an opt-out, and provide an easy link for us to reply to those security report emails.

> While it follows common patterns, a decorator for task registration does affect usability compared to just sending the callable, because it requires that the modules be imported in order for the tasks to be found. This can be done explicitly or by some auto-discovery magical convention, and while you may prefer those patterns they aren't without trade-offs (ones that I would prefer to avoid).

Django is opinionated and already applies the auto-discovery pattern for several types of resources; most relevant examples being [Models](https://github.com/django/django/blob/main/django/apps/config.py#L262-L269) and [management commands](https://github.com/django/django/blob/main/django/core/management/__init__.py#L76-L78). Adding task auto-discovery to AppConfig with a `def import_tasks(self)` would be a natural fit. 

This AppConfig auto-discovery opens the opportunity for a built-in management commands to inspect and run/enqueue tasks. I've seen this exact cron calling a management command that enqueues a celery task pattern way too many times in my career. It would be nice to remove that boilerplate code because people will repeat that pattern if cron is not included.

PeterJCLaw: > Django is opinionated and already applies the auto-discovery pattern for several types of resources

FWIW, [DLQ](https://github.com/thread/django-lightweight-queue) takes this approach -- it auto-loads `tasks.py` files in apps to discover tasks which register themselves via a decorator.

draft/0000-background-workers.rst
PeterJCLaw: A separate thought on naming, which I think someone mentioned in a different thread but I can't immediately find -- I think it would be useful to separate the ideas of:
- a function which can be enqueued & called later
- the representation of such an enqueueing which is pending being run

Currently these seem to both be called "tasks", which potentially leads to confusion. One spelling for splitting this out might be tasks (definitions) & jobs (instances) respectively[^1].

[^1]: this terminology stolen from [DLQ](https://github.com/thread/django-lightweight-queue).

draft/0000-background-workers.rst
PeterJCLaw: FWIW: [DLQ](https://github.com/thread/django-lightweight-queue) doesn't store either results or statuses. In its model, tasks are expected to persist their outputs themselves (usually to the database) and task instances don't exist at all once they're complete.

draft/0000-background-workers.rst
ryanhiebert: In some of the latest updates he's referred to the task definition as a "task function", as opposed to the task which is a "job" in your definition above. I'm not sure we'll find terminology everyone is happy with, so I've lived with the ambiguity, although I did make an oblique suggestion that we need a glossary at the end of [this comment](https://github.com/django/deps/pull/86#discussion_r1501960629).

draft/0000-background-workers.rst
RealOrangeOne: Given the type of it has to be `Any`, there's little a developer can do with the value sensibly. I agree, let's drop it, and rely on someone introspecting the `Task` if they need to go deeper.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,383 @@
...
+
+   @task
+   def do_a_task(*args, **kwargs):
+      pass
+
+   # Submit the task function to be run
+   task = enqueue(do_a_task)
+
+   # Optionally, provide arguments
+   task = enqueue(do_a_task, args=[], kwargs={})
```
benjaoming, line 232: Thanks for the amazing work :heart_eyes: 

Since `task` in this case is both a decorator and then redefined as variable, maybe it's preferable to rename the decorator to what it does, i.e. "register the task". Like Wagtail hooks, template tags etc.

That made me wonder if there can be more than one registry of tasks with different backends and/or route to different queues.

```
from django.tasks import register_task

@register_task()
def my_task(...)

@register_task(runner="external_cluster")
def my_other_task(...)
```

RealOrangeOne: I intentionally didn't add any data to the `task` decorator, as its only function is to allow-list registrations. I don't want to add any arguments to it which assume how the backends are configured, as a library maintainer may assume a `external_cluster` queue, which won't actually exist. Instead, doing it at the call site makes things much simpler.

benjaoming: @RealOrangeOne feeding the argument to `register_task` was an example. But the _real_ suggestion was to rename to decorator :)

> its only function is to allow-list registrations

I think that "register the task" is its scope. But indeed, as you mention, there are different ways to define this. I think it's fine to leave kwargs for future discussions. I know it's a door that may invite in endless suggestions that will make the DEP harder to finish and implement :+1:

benjaoming: The decorator could also be called `allow_as_task`, if the narrower scope is truly more desirable.

draft/0000-background-workers.rst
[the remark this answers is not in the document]
adamchainz: Python’s convention is that naive datetimes are *local time*. We should not use them to represent UTC times, as that could lead to confusion. So, we should enforce aware datetimes.

adamchainz:
I would rather we don’t use Pickle for argument serialization, and instead use `json`. There are too many problems with pickle:

1. **Protocol version changes** between Python versions complicate upgrading, since you need to lose tasks or temporarily try both versions.
2. **Enqueueing data rather than references** - That is, enqueueing a task with a model instance, rather than its ID. If you save the queued model instance within the task, it erases any intermediate changes in the database. I’ve seen this problem so many times that I put it first in my post [“Common Issues Using Celery (And Other Task Queues)”](https://adamj.eu/tech/2020/02/03/common-celery-issues-on-django-projects/#enqueueing-data-rather-than-references).
3. **Security problems** - per the big red box in [the pickle docs](https://docs.python.org/3/library/pickle.html). Yes, the task storage should be trusted, just like cache storage, but if we can just avoid pickle altogether there will never be an issue.
4. **Slower performance** - pickle makes it easy to put lots of data into a task argument when you only need a bit, such as a whole model instance when you only need an ID. It’s also not the fastest to deserialize because it runs a whole stack machine.

The serialization should be swappable, perhaps by subclassing and replacing the serialize/deserialize methods. That way users can use pickle or whatever if they really need to.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,397 @@
...
+
+   # Override the priority defined by the `Task`
+   task = enqueue(do_a_task, priority=10)
```
adamchainz, line 241: I’d really prefer we only provide an API like:

```python
task.enqueue(*args, **kwargs)
```

This would allow type-checking, as the type hints should be able to preserve the signature through a `ParamSpec` mapped to `args` and `kwargs`.

For extra arguments, we can use a double call syntax like:

```python
task.enqueue_with(priority=10)(*args, **kwargs)
```

ryanhiebert: I agree, and your suggested API looks good to me. Since the task decorator is required, avoiding the top-level enqueue method seems reasonable. We should avoid APIs that cripple type checking. Plus this calling API is a more ergonomic option for simple uses.

RealOrangeOne: `ParamSpec` support has been vaguely on my list to ensure. The main reason for this API is to avoid tasks shadowing paramters like `priority`, and it not being clear. 

I'd assumed it was possible to type anything with a `ParamSpec.args` / `ParamSpec.kwargs`, but turns out it only likes being used on `*args` and `*kwargs` (at least as far as `mypy` is concerned).

The double call syntax might be a little intimidating and odd, especially for new developers. But there might be an ORM-style pattern which works:

```python
task.with(priority=10).enqueue(*args, **kwargs)
task.with(priority=10).enqueue(*args, **kwargs)
task.priority(10).enqueue(*args, **kwargs)
```

This is a notable change from a global function to being something which hangs off a `Task`, but that's probably ok, and may even make calling tasks much more obvious and readable.

RealOrangeOne:
> The serialization should be swappable

I had intentionally not included support for this. If a given backend doesn't support pickle-based serialization, shoe-horning it in can be quite difficult and unpredictable. If a library maintainer also assumes pickle serialization, their library won't work in environments where the pickle constraints do matter. And that's before we think of a transition path.

Going with JSON (at least at first) has a few benefits:

- It avoids the pickle issues
- If it can be JSON serialized, it can be pickled (the inverse isn't true)
- It bypasses a number of odd behaviours where someone accidentally passes something to a task which shouldn't be (ie it isn't thread safe or depends on external state like `NamedTemporaryFile`).
- Keeping the format JSON serializable can be helpful when integrating with external monitoring tools

I'm not opposed to swappable serializers in future, but JSON only for the initial pass is probably the way to go. Being limiting and then expanding support is going to be better for adoption than being vague or potentially limiting in future.

draft/0000-background-workers.rst
RealOrangeOne: Refactoring in this way has lead to a much cleaner API I'm much happier with - thanks! The types should now flow end-to-end if you use the right APIs.

prophile:
> The serialization should be swappable, perhaps by subclassing and replacing the serialize/deserialize methods. That way users can use pickle or whatever if they really need to.

`json` as default serialization sidesteps some problems and avoids the issues around serializing instances, but not being able to serialize a `datetime` in an argument by default is quite unfortunate. I don't know whether it's worth a default workaround though.

carltongibson (commented):
Hey @RealOrangeOne. 

This is looking good. A very gentle read, and really quite compelling. 🏅 

With the (WIP) reference implementation coming  along https://github.com/RealOrangeOne/django-core-tasks it's very exciting. (There are no Issues there, so I'll comment here: would it be worth adding an `examples/` folder with a simple app setup, so that first pass is not much more than and `git clone`, a `pip install`, and a `./manage.py ... ` or two?)

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,407 @@
...
+      pass
+
+
+A ``Task`` can only be created for globally-importable callables. The task will be validated against the backend's ``is_valid_task`` callable during construction.
+
+If a task doesn't define a backend, it is assumed it will only use the default backend.
+
+``@task`` may be used on functions or coroutines. It will be up to the backend implementor to determine whether coroutines are supported. Support for coroutine tasks can be determined with the ``supports_coroutine_tasks`` method on the backend. In either case, the function must be globally importable.
+
+Task arguments must be JSON serializable, to avoid compatibility and versioning issues. Complex arguments should be converted to a format which is JSON-serializable.
```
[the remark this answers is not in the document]
RealOrangeOne: I'd love to be able to use `DjangoJSONEncoder`, to add a few more data types, but sadly those operations won't work the other way around, as there's no deserialization support, which would result in the types being completely different.

We could flag the data type alongside, but that requires related metadata carried separately, or a proprietary format for the arguments.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,407 @@
...
+``BaseTaskBackend`` will provide ``a``-prefixed stubs for ``enqueue`` and ``get_task`` using ``asgiref.sync_to_async``.
+
+``is_valid_task`` determines whether the provided ``Task`` is valid for the backend. This can be used to prevent coroutines from being executed, or otherwise validate the callable. If a backend receives a task which is not valid (ie ``is_valid_task`` returns ``False``), it should raise ``InvalidTaskError``. The base implementation of ``is_valid_task`` will validate:
+
+- Is the task's function a valid, globally-importable callable?
+- Is the task allowed to be run on the current backend?
+
+If a backend cannot support deferred tasks (ie passing the ``run_after`` argument), it should raise ``InvalidTaskError``. The ``supports_defer`` method can be used to determine whether the backend supports deferring tasks.
+
+Django will ship with 3 implementations:
```
ngnpope, line 89: It's better to avoid explicit counts altogether:

```suggestion
Django will ship with the following implementations:
```

draft/0000-background-workers.rst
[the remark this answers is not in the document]
carltongibson: Take `datetime` as an example. From JSON I'm going to need to take a string, and then convert it in my task function. But I might want to also accept an actual `datetime`, since likely I've got one in hand. If I could specify the encoder, I could at least use `DjangoJSONEncoder` myself, rather than having to manual convert that `datetime` first. 🤔

ngnpope (commented):
I've had these comments floating around as pending for ages and never got around to posting them. Doing so now - I did have a quick look and tweaked them, but sorry if they're outdated or have been discussed/decided on already.

draft/0000-background-workers.rst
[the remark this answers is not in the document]
PeterJCLaw: While this is true, the generic argument of some additional encoding/decoding needing to happen applies to all sorts of things. Model or enum instances are the main ones which come to mind beyond those already mentioned.

My experience has been that adding an API function (often named `enqueue_foo`) which wraps a task in a shallow manner and which does take the Python instances typically works fairly well for this sort of thing. This approach also means that the API can provide higher level type safety, which isn't something that an encoder based solution could trivially offer.

I'm not against adding an encoder option later (especially if we can ensure round-tripping of the data), rather highlighting that there's an easy alternative for the meanwhile which alleviates some of the downsides to deferring this.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,407 @@
...
+      queue_name: str | None
+      """The name of the queue the task will run on """
+
+      backend: str
+      """The name of the backend the task will run on"""
+
+      run_after: datetime | None
+      """The earliest this task will run"""
+
+      def using(self, priority: int | None = None, queue_name: str | None = None, run_after: datetime | None = None) -> Self:
```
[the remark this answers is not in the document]
RealOrangeOne: I suspect the "Run in X minutes" use case is far more common than "run at X time", so I agree it makes sense to support, even just as shorthand.

It'll be implied the `timedelta` will be relative to `timezone.now()`, which _hopefully_ won't lead to any foot-gun scenarios when the user is balancing multiple timezones for multiple users.

draft/0000-background-workers.rst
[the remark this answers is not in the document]
RealOrangeOne: I'm not opposed to that, but I don't think it should be day _one_. Definitely closely followed, but developing and maintaining 2 backends inside Django as part of the initial release will likely delay its inclusion quite substantially. A database backend should work for the majority of users, and the rest will either need to use a 3rd-party backend, or wait for the first-party Redis support.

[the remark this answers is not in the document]
orf: The advantage to having a redis backend as well is that we can ensure compatibility: we don’t want to be in a situation where we are baking in assumptions about the initial backend into the API, and find that this doesn’t work with, is not performant on or isn’t compatible with Redis in some way.

I’m not sure what those might be, and honestly I’m not sure it’s an issue right now - we don’t want to delay the implementation. However, perhaps a quick, hacky non-production PoC with Redis or another broker might be worth trying, just to see if there are any unexpected issues we run into relating to the API surface?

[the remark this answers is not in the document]
RealOrangeOne: > a quick, hacky non-production PoC with Redis or another broker might be worth trying

I think this is the approach to take. A mixture of at least _trying_ an alternative implementation, but also either working with other library developers, or getting their help with API design should get all the teething issues out of the way early.

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,420 @@
...
+The field of background tasks is vast, and attempting to implement everything supported by existing tools in the first iteration is futile. The following functionality has been considered, and deemed explicitly out of scope of the first pass, but still worthy of future development:
+
+- Completion / failed hooks, to run subsequent tasks automatically
+- Bulk queueing
+- Automated task retrying
+- A generic way of executing task runners. This will remain the responsibility of the underlying implementation, and the user to execute correctly.
+- Observability into task queues, including monitoring and reporting
+- Cron-based scheduling
+- Task timeouts
+- Swappable argument serialization (eg `pickle`)
```
olivierdalang, line 414: Maybe we can include these in this list (which I see as potentially important and which I don't see discussed in the DEP)

- Logging (besides the result, the task results could capture logs, stdout and stderr)
- Realtime output (for long running tasks, having the ability to get some kind of feedback before it finishes would for instance allow to implement progress bars)
- Non-django tasks (not 100% clear if the proposed approach already supports enqueuing non-django tasks, but if not, having a worker agnostic background task API would open quite some doors for distributed architectures)

draft/0000-background-workers.rst
```diff
@@ -0,0 +1,420 @@
...
+
+If a backend cannot support deferred tasks (ie passing the ``run_after`` argument), it should raise ``InvalidTaskError``. The ``supports_defer`` method can be used to determine whether the backend supports deferring tasks.
+
+Django will ship with the following implementations:
+
+ImmediateBackend
+   This backend runs the tasks immediately, rather than offloading to a background process. This is useful both for a graceful transition towards background workers, but without impacting existing functionality.
+
+DatabaseBackend
+   This backend uses the Django ORM as a task store. This backend will support all features, and should be considered production-grade.
```
olivierdalang, line 98: As you explained, this will suit most users. Could you describe just a little bit more how you'd see that implemented (as I didn't see it yet in your POC implementation) ?
I guess it would be an optional contrib module providing a management command to run the workers, and probably some admin integration to see task results ?

olivierdalang (commented):
As an author of one of these too many packages dealing with background tasks (django-toosimple-q), I can't wait to see this land in Django :-) Thanks so much for working on this !

draft/0000-background-workers.rst
RealOrangeOne: I've intentionally not put implementation details like that in here, as they're fairly external. But yes, the idea would be a management command to run the worker, a DB model to store the data itself, and likely an admin interface for debugging.

draft/0000-background-workers.rst
RealOrangeOne: - Capturing stdout / stderr is probably an antipattern anyway - collect metrics like that manually
- As part of the initial interface, that's definitely not happening without manual implementation. Depending on how granular the progress is, it could end up a performance bottleneck
- This one has bugged me for a while - there's not a huge amount in here that's Django specific, so it _could_ be extracted out to Python itself. However, I'd rather let the backends deal with that themselves. If someone wants to maintain a package which can let arbitrary Python apps schedule tasks using say the built-in DB queue, I'm not opposed to that at all.

draft/0000-background-workers.rst
RealOrangeOne: Since this discussion, I've made quite a few changes to the API, including adding an explicit decorator to create a `Task` object based on a function (which in turn both validates the function is globally importable, and ensures only registered functions can be used as tasks).

I'm not super familiar with how Celery "marks" a task behind the scenes, and then validates it. If it just needs to store a function reference somewhere, that's easy enough to implement with the current calling pattern. If it needs the globally importable function to be of a given type, that might also be doable with a little more work. 

@ryanhiebert as you've clearly used Celery more than I, I'm interested in your thoughts!

ryanhiebert: it needs it to be of a specific type (a subclass of celery.Task), and registered with the (usually global) celery app, I believe.

apollo13:
Overall the proposal looks good, but there is one thing that I can't wrap my head around it yet. Let's assume you are writing a third-party application and want to use this task framework. Which *backend*/*queue* names are you putting into the `@task` decorator? If you leave it at `default` then all of a sudden most third party libraries will end up on the same backend. Using a unique backend name for your third party library requires the user to configure multiple backends which all in the worst case have an expensive to setup connection. In an ideal world as a consumer of the library I'd like to be able to route where the tasks go without having to fork the library.

Two approaches come to mind (there might be more and even saner approaches):

 * Use the `queue_name` of the task to search for a backend that serves this queue?
 * Add `aliases` to the backend definition so that one could configure something like this:

```python
   TASKS = {
      "high_prio": {
         "BACKEND": "django.tasks.backends.ImmediateBackend",
         "QUEUES": [],
         "OPTIONS": {},
         "ALIASES": ["default", "third_party_app1"] # See here
      },
      "low_prio": {
         "BACKEND": "django.tasks.backends.ImmediateBackend",
         "QUEUES": [],
         "OPTIONS": {},
         "ALIASES": ["third_party_app2"] # and here…
      }
   }
```

apollo13:
I think the points raised by @prophile in `Transactions in ImmediateBackend` need some more thoughts. Even if we want *fast* tests we shouldn't stray to much in semantics from a real deployment. This most likely means that we have to execute tasks from the immediate backend in a separate database connection.

Additionally we should probably add a flag to the database backend so it can operate in two ways: one where it joins the existing transaction on `enqueue` and one where it utilizes `on_commit` if a transaction is open. 

The database backend part doesn't have to be part of the initial implementation, but we really need to get the semantics of the immediate backend right.

RealOrangeOne:
> Which backend/queue names are you putting into the `@task` decorator?

Unless you have a good reason, just omit it. If you want users to be able to configure the attribute themselves, you can define your own setting and let the user configure it there. I do like the `aliases` idea, as a way of easily allowing a user to route a task to a given queue, but a setting feels a bit less _magic_ to me.

I imagine `TASKS` being very small, generally just a single entry - much like you would have in `DATABASES` for most deployments. High / low priority should be handled either with the `priority` attribute, or using queue names if the intention is more around dedicated capacity / specialised hardware.

RealOrangeOne:
> Transactions in ImmediateBackend

These are the semantics which concern me about an initial implementation. I suspect configuration is the way to go here, as there isn't one obvious answer (`on_commit` might make sense for a database-based backend, but is radically different to an external, eg Redis, backend).

apollo13:
> > Which backend/queue names are you putting into the `@task` decorator?
> 
> Unless you have a good reason, just omit it. If you want users to be able to configure the attribute themselves, you can define your own setting and let the user configure it there. I do like the `aliases` idea, as a way of easily allowing a user to route a task to a given queue, but a setting feels a bit less _magic_ to me.

I didn't consider a setting here; but I agree that it is a good workaround for now and we can add `aliases` later on if needed.

> > Transactions in ImmediateBackend
> 
> These are the semantics which concern me about an initial implementation. I suspect configuration is the way to go here, as there isn't one obvious answer

I guess the immediate answer would be "behave like other backends would" :D But once one tested that the backends behave properly it might make sense to switch to "I know this code works, now run fast so my tests are quicker and better to test".

carltongibson:
@RealOrangeOne As per the Forum thread, I think you're good to move this to the `accepted` folder, and then I think we can merge.

RealOrangeOne:
Done - Over to you @carltongibson! :partying_face:

carltongibson (approved):

orokusaki:
FWIW, I think this is a mistake.

I suppose it's too late now, but I'll still put my thoughts here for posterity:

I would like to start by suggesting that the xkcd in the OP, ironically, does not come across ironically. in fact, it describes exactly what we are talking about here. There are **many** task queue systems for Python and Django.

Queueing systems solve more than just "long-running tasks". RQ works well, same for Celery, too, along with a number of others. Running tasks is not intrinsic to the web, because it does not relate specifically to web functionality, like HTTP requests, headers, basically anything at all. It is exactly not related to Web frameworks whatsoever. Web protocols are inherently opinionated, which makes a lot of the choices in a web framework obvious and simple. Task queueing is not obvious or simple. I went to look at the reference implementation and was immediately put off by the need to use a decorator to turn each task function into a task API. I would also like to suggest that, based on my prima facie reading of @prophile big comment above, the fundamental machinery involved it's not fully understood by the parties involved herein. There are many reasons to do many things very differently than  what is being suggested here, and there isn't any reason to debate them, because we've done without this debate for 19 years and people had a choice and the ecosystem thrived.

Most projects that I work on need a task queue. You know what else all the apps I work on need? Electricity and a server. Should Django build some power plants and rent co-lo space? I mean, if we're talking about adding things that are wholly unrelated to the web simply "because batteries included", I can't think of a better way to continue fulfilling that promise. Not everybody uses the authentication system, but we have that. Given that, one could make an argument that we should also add payment processing, since many many apps use it. We can really stay ahead of the pack if we start thinking about crypto support. Let's turn this into like one big Monday + Zapier + CoinBase style app + a ski resort closer to Florida!

Task queues have been around since the beginning, and somehow Django has managed without building them in for **19 years**. One could make the argument that databases are also completely unrelated to the web. However, these are trade-offs that were made many many moons ago, because data, logic, display are fundamental requirements for doing anything meaningful on the web.

in summary, this reminds me of Andrew Godwins attempt to build Channels into Django. I was the one who made a stink about that as well, on the Google groups list, for the same reason; this feels like a high school project that is going to badly shape the future of something very big and important. Looking back, aren't we glad we didn't merge that in? There's so many ways to do what channels does.

if somebody wants to get involved and help the framework, why not focus on async and/or performance optimization? A vast majority of the market share Django has lost over the years is specifically due to the perceived performance and/or async benefits of using micro frameworks. I didn't run a poll, but I've been making software in Django since 0.96, and I've worked with a lot of people, and known many who have jumped ship for micro frameworks for these reasons. I'm sure everyone of you knows somebody who fits that bill.

accepted/0014-background-workers.rst
```diff
@@ -0,0 +1,422 @@
...
+
+Settings
+---------
+
+.. code:: python
+
+   TASKS = {
+      "default": {
+         "BACKEND": "django.tasks.backends.ImmediateBackend",
+         "QUEUES": []
```
PeterJCLaw, line 353: nit
```suggestion
         "QUEUES": [],
```

accepted/0014-background-workers.rst
```diff
@@ -0,0 +1,422 @@
...
+         ...
+
+
+A ``TaskResult`` will cache its values, relying on the user calling ``refresh`` to reload the values from the task store. An ``async`` version of ``refresh`` is automatically provided by ``TaskResult`` using ``asgiref.sync_to_async``.
+
+A ``TaskResult``'s ``status`` must be one of the following values (as defined by an ``enum``):
+
+:NEW: The task has been created, but hasn't started running yet
+:RUNNING: The task is currently running
+:FAILED: The task failed
```
PeterJCLaw, line 224: Please could we clarify what's meant by "failed" here? Is this timed out? Threw a particular exception? Threw any exception? Returned an unexpected result? Returned no result where one was expected?
Is this something the backends are expected to create rules for themselves?

I suspect that "threw any exception" is probably the simplest and may be what's intended, though it would be good to say as much IMO.

accepted/0014-background-workers.rst
```diff
@@ -0,0 +1,422 @@
...
+:COMPLETE: The task is complete, and the ``result`` is accessible
+
+If a backend supports more than these statuses, it should compress them into one of these.
+
+For convenience, calling a ``TaskResult`` will execute the task's function directly, which allows for graceful transitioning towards background tasks:
```
PeterJCLaw, line 229:
```suggestion
For convenience, calling a ``Task`` will execute the task's function directly, which allows for graceful transitioning towards background tasks:
```
I assume this is what's intended here?

PeterJCLaw:
Sorry to add a bunch of comments right after this has merged. I've been following this PR with interest though hadn't realised that it had progressed to a near final state -- I would have tried to re-review it more completely prior to now if I had.

accepted/0014-background-workers.rst
RealOrangeOne: Yes, yes it is :facepalm:

accepted/0014-background-workers.rst
```diff
@@ -0,0 +1,422 @@
...
+.. code:: python
+
+   from django.tasks import task
+
+   @task()
+   def do_a_task(*args, **kwargs):
+      pass
+
+   # Calls `do_a_task` as if it weren't a task
+   do_a_task()
```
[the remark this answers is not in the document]
RealOrangeOne: During implementation, I his this foot-gun too. My proposed solution is a `.call` method (and `.acall`) instead of implementing `__call__`. I do know some task libraries use `__call__` to enqueue a task, so I agree this could be confusing.

I'd also thought of additionally having `__call__ enqueue a task, but that felt like the wrong way forward. "There should be one-- and preferably only one --obvious way to do it."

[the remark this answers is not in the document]
PeterJCLaw: What benefit does a `Task.call`/`Task.acall` method offer when `Task.func` already exists? (Assuming I'm reading the DEP right about that member)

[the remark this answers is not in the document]
orf: In the context of beginners, `.func(…)` doesn’t really convey much intent. `.call()` is a bit clearer. `.func` also gives you the underlying function object, it’s possible we want some extra logic in `enqueue` and `call`.

RealOrangeOne:
I've now opened up [discussions](https://github.com/RealOrangeOne/django-tasks/discussions) on the [in-progress implementation](https://github.com/RealOrangeOne/django-tasks), and pre-populated it with the "Future iterations" items. 

I suggest we move further discussions on the DEP there. After the initial influx of comments and feedback there, I'll backport the relevant changes to the DEP in a separate PR (hence the distinction between "accepted" and "final"). Please do raise any and all comments / questions there (and PRs too)!

accepted/0014-background-workers.rst
adamchainz: Done in #90.

PR #88, opened 2024-05-29
DEP 0017 Content Type Parsing

smithdc1:
Hi @carltongibson 👋 

I've made a start on writing up a DEP following the discussion on the [forum](https://forum.djangoproject.com/t/request-for-steering-council-vote-on-modernising-the-request-object/26816). This is based upon previous discussions both on the mailing list, the forum and the linked PRs. 

There's a couple of "TODO"s which need a bit more thought. But thought worth sharing where I have got to. 

I hope that I have presented a fair and unbiased proposal of where we are and the concerns raised.

### DIFF
--- a/draft/0015-content-types.rst
+++ b/draft/0015-content-types.rst
@@ -26,7 +26,7 @@ Parsed data from an ``HttpRequest`` is accessed via its ``POST`` attribute. It w
 
 While introducing a new name for ``POST`` it is proposed that the names for the other attributes are modernized with an equivalent behaviour.
 
-The existing ``GET``, ``POST``, ``META``, and ``FILES`` attributes will be maintained for backwards compatibility, and the behaviour (specifically of POST) will remain unchanged.
+The existing ``GET``, ``POST``, ``COOKIES``, ``META``, and ``FILES`` attributes will be maintained for backwards compatibility, and the behaviour will remain unchanged.
 
 Specification
 =============
--- a/draft/0015-content-types.rst
+++ b/draft/0015-content-types.rst
@@ -1,0 +1,149 @@
+DEP 0015: Content type aware parsing and modernization of the HttpRequest API.
+
+DEP
+
+    0015
+
+Author
+
+    David Smith
+
+Implementation Team
+
+    TBC
+
+Shepherd
+
+    TBC
+
+Status
+
+    Draft
+
+Type
+
+    Feature
+
+Created
+
+    2024-03-25
+
+Last-Modified
+
+    2024-05-29
+
+Table of Contents
+
+
+Abstract
+========
+
+Currently Django can parse requests for ``application/x-www-form-urlencoded`` and ``multipart/form-data`` types. Other types, such as JSON are currently returned as a string.
+
+This DEP proposes to add configurable content type parsers to allow parsing of additional content types. It is proposed that Django will include a parsing of JSON, and appriate hooks to allow users to add custom parsers for other content types.
+
+Parsed data from an ``HttpRequest`` is accessed via its ``POST`` attribute. It would be a breaking change if Django were to start parsing content types where currently a string is returned. To avoid introducing a breaking change it is proposed that a new ``data`` attribute is added for the new behaviour.
+
+While introducing a new name for ``POST`` it is proposed that the names for the other attributes are modernized with an equivalent behaviour.
+
+The existing ``GET``, ``POST``, ``META``, and ``FILES`` attributes will be maintained for backwards compatibility, and the behaviour (specifically of POST) will remain unchanged.
+
+Specification
+=============
+
+Django shall provide parsers for ``application/x-www-form-urlencoded``, ``multipart/form-data`` and ``application/json`` content types.
+
+A parser should be a class which is instantiated with a request and have two methods.
+
+- ``can_handle()`` should accept one argument being the ``media_type`` attempting to be parsed and returns a boolean to indicate if this parser can parse the given media type.
+- ``parse()`` accepts one argument being the ``data`` to parse and returns the parsed data.
+
+By default all requests shall be parsed by Django's default parsers. But to ensure backward compatability the new behaviour shall only be available when accessing the parsed response via the new ``data`` attribute.
+
+The new ``data`` attribute should parse requests in as close a way as possible to the current ``POST`` attribute. In addition it shall:
+
+* Parse ``application/JSON....`` types.
+* For ``multipart/form-data`` types each part will be parsed with the appropriate parser, if available.
+* Raise an ``UnsupportedMediaType`` (415) error if an unsupported content type is attempted to be parsed.
+
+The new data attribute is therefore not 100% equivalent to POST since it will parse JSON (and other data types as parsers for those are configured) where POST would return a string.
+
+Custom parsers for additional content types shall be supported. To allow this it is proposed that ``HttpRequest`` adds a new property which returns a list of parsers to be used when attempting to parse a request.
+The list of parsers to be used can be set on the request, but must be done before ``data`` or ``FILES`` is accessed. This could be done in a middleware or in a view. For example::
+
+    def index(request):
+        request.parsers = [MyCustomParser(), FormParser(), ...]
+        ...
+
+To mitigate backward compatability concerns the new behaviour shall be accessed using the new ``data`` attribute. This introduces new, lower case names which better reflect the behaviour of the function.
+At the same time it is proposed that the following attributes shall be added to modernise the other names:
+
+* query_params (for GET)
+* files (for FILES)
+* meta (for META)
+* cookies (for COOKIES)
+
+The behaviour of renamed attributes shall be 100% compatible with the existing attributes.
+
+# TODO
+
+A separate [ticket 17235](https://code.djangoproject.com/ticket/17235) is accepted which proposes to leave FILES as immutable.
+Do we need to mention this here? The new behaviour could be on the lower cased name only?
+
+Motivation
+==========
+
+This DEP is needed to add configurable content type parsing to Django. While this is likely to be well received the proposal also suggests adding new aliases which is more controversial.
+
+The motivation to improve the names is that ``request.GET`` and ``request.POST`` are misleadingly named:
+
+* ``GET`` contains the URL parameters and is therefore available whatever the request method. This often confuses beginners and "returners" alike.
+
+* ``POST`` contains form data on ``POST`` requests, but not other kinds of data from ``POST`` requests. It can confuse users who are posting JSON or other formats.
+
+Additionally both names can lead users to think e.g. "if request.GET:" means "if this is a GET request", which is not true.
+
+The CAPITALIZED naming style is similar to PHP's global variables $_GET, $_POST, $_FILES etc. (https://www.php.net/manual/en/reserved.variables.get.php ). It stands out as unpythonic, since these are instance variables and not module-level constants (as per PEP8 https://www.python.org/dev/peps/pep-0008/#constants).
+
+However, with ``HttpRequest`` being such a core part of Django renaming these will cause a large amount of churn. The change to the documentation will be significant and many existing tutorials, blog posts and books by authors in the community would require updating to reflect the new, recommended appraoch.
+As such it is proposed that the new names are not immediately deprecated.
+
+# TODO
+
+What would we like to say about a deprecation path? What would have to be true for it to even be considered?
+
+See mailing list conversation [1]
+
+[1] https://groups.google.com/g/django-developers/c/Kx8BfU-z4_E/m/gJBuGeZTBwAJ
+
+Rationale
+=========
+
+The main objection received by the community is the renaming of the attributes. This causes a lot of churn in documentation to rename attributes where the behaviour of these is equivielent.
+
+Other options are:
+
+- Leave additional content type parsing to 3rd party packages, e.g. DRF
+- Introduce content type parsing and only add the new ``data`` attribute.
+
+The new names for unchanged attributes is proposed as it's considered this a worthwhile improvement in its own right and introduces consistent naming across ``HttpRequest`` attributes. That is, without renaming the change only the new ``data`` attribute would be an outlier.
+
+Backwards Compatibility
+=======================
+
+This DEP is designed to be backward compatible. The existing ``GET``, ``POST``, ``META``, and ``FILES`` attributes will be maintained for backwards compatibility, and (to emphasise again) the behaviour (specifically of POST) will remain unchanged.
+
+This is similar to the way the headers property was added, whilst maintaining the older dictionary style lookup.
+
+Reference Implementation
+========================
+
+There are currently two PRs which are work towards implementation of this DEP.
+
+* Addition of content type parsing https://github.com/django/django/pull/17546
+* Modernization of Request Object attribute names https://github.com/django/django/pull/17624
+
+Copyright
+=========
+
+This document has been placed in the public domain per the Creative Commons CC0 1.0 Universal license (http://creativecommons.org/publicdomain/zero/1.0/deed).
--- a/draft/0015-content-types.rst
+++ b/draft/0015-content-types.rst
@@ -24,7 +24,13 @@ This DEP proposes to add configurable content type parsers to allow parsing of a
 
 Parsed data from an ``HttpRequest`` is accessed via its ``POST`` attribute. It would be a breaking change if Django were to start parsing content types where currently a string is returned. To avoid introducing a breaking change it is proposed that a new ``data`` attribute is added for the new behaviour.
 
-While introducing a new name for ``POST`` it is proposed that the names for the other attributes are modernized with an equivalent behaviour.
+While introducing a new name for ``POST`` it is proposed that the names for the other attributes are modernized with an equivalent behaviour:
+
+* ``GET`` -> ``query_params``
+* ``POST`` -> ``form_data``
+* ``COOKIES`` -> ``cookies`
+* ``META`` -> ``meta``
+* ``FILES`` -> ``files``
 
 The existing ``GET``, ``POST``, ``COOKIES``, ``META``, and ``FILES`` attributes will be maintained for backwards compatibility, and the behaviour will remain unchanged.
 
--- a/draft/0015-content-types.rst
+++ b/draft/0015-content-types.rst
@@ -22,13 +22,13 @@ Currently Django can parse requests for ``application/x-www-form-urlencoded`` an
 
 This DEP proposes to add configurable content type parsers to allow parsing of additional content types. It is proposed that Django will include a parsing of JSON, and appriate hooks to allow users to add custom parsers for other content types.
 
-Parsed data from an ``HttpRequest`` is accessed via its ``POST`` attribute. It would be a breaking change if Django were to start parsing content types where currently a string is returned. To avoid introducing a breaking change it is proposed that a new ``data`` attribute is added for the new behaviour.
+Parsed data from an ``HttpRequest`` is currently accessed via its ``POST`` attribute. It would be a breaking change if Django were to start parsing content types where currently a string is returned. To avoid introducing a breaking change it is proposed that a new ``data`` attribute is added for the new behaviour.
 
 While introducing a new name for ``POST`` it is proposed that the names for the other attributes are modernized with an equivalent behaviour:
 
 * ``GET`` -> ``query_params``
 * ``POST`` -> ``form_data``
-* ``COOKIES`` -> ``cookies`
+* ``COOKIES`` -> ``cookies``
 * ``META`` -> ``meta``
 * ``FILES`` -> ``files``
 
@@ -61,20 +61,19 @@ The list of parsers to be used can be set on the request, but must be done befor
         request.parsers = [MyCustomParser(), FormParser(), ...]
         ...
 
-To mitigate backward compatability concerns the new behaviour shall be accessed using the new ``data`` attribute. This introduces new, lower case names which better reflect the behaviour of the function.
+To mitigate backward compatability concerns the new behaviour shall be accessed using the new ``data`` attribute.
+
 At the same time it is proposed that the following attributes shall be added to modernise the other names:
 
-* query_params (for GET)
-* files (for FILES)
-* meta (for META)
-* cookies (for COOKIES)
+* ``GET`` -> ``query_params``
+* ``POST`` -> ``form_data``
+* ``COOKIES`` -> ``cookies``
+* ``META`` -> ``meta``
+* ``FILES`` -> ``files``
 
-The behaviour of renamed attributes shall be 100% compatible with the existing attributes.
+The behaviour of renamed attributes shall be 100% compatible with the existing attributes. To ease migration to the new names django-upgrade can be used to automate refactoring code. Some work to show this is possible is available [1].
 
-# TODO
-
-A separate [ticket 17235](https://code.djangoproject.com/ticket/17235) is accepted which proposes to leave FILES as immutable.
-Do we need to mention this here? The new behaviour could be on the lower cased name only?
+[1] https://github.com/smithdc1/django-upgrade/commit/c043761cb2fa00e97c3ea205be7e79fcec6f075d
 
 Motivation
 ==========
@@ -94,13 +93,9 @@ The CAPITALIZED naming style is similar to PHP's global variables $_GET, $_POST,
 However, with ``HttpRequest`` being such a core part of Django renaming these will cause a large amount of churn. The change to the documentation will be significant and many existing tutorials, blog posts and books by authors in the community would require updating to reflect the new, recommended appraoch.
 As such it is proposed that the new names are not immediately deprecated.
 
-# TODO
+See mailing list conversation [2]
 
-What would we like to say about a deprecation path? What would have to be true for it to even be considered?
-
-See mailing list conversation [1]
-
-[1] https://groups.google.com/g/django-developers/c/Kx8BfU-z4_E/m/gJBuGeZTBwAJ
+[2] https://groups.google.com/g/django-developers/c/Kx8BfU-z4_E/m/gJBuGeZTBwAJ
 
 Rationale
 =========
@@ -133,3 +128,4 @@ Copyright
 =========
 
 This document has been placed in the public domain per the Creative Commons CC0 1.0 Universal license (http://creativecommons.org/publicdomain/zero/1.0/deed).
+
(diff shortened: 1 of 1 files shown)


draft/0015-content-types.rst
```diff
@@ -0,0 +1,149 @@
...
+
+The main objection received by the community is the renaming of the attributes. This causes a lot of churn in documentation to rename attributes where the behaviour of these is equivielent.
+
+Other options are:
+
+- Leave additional content type parsing to 3rd party packages, e.g. DRF
+- Introduce content type parsing and only add the new ``data`` attribute.
+
+The new names for unchanged attributes is proposed as it's considered this a worthwhile improvement in its own right and introduces consistent naming across ``HttpRequest`` attributes. That is, without renaming the change only the new ``data`` attribute would be an outlier.
+
```
adamchainz, line 130: Maybe we can also add a note that django-upgrade could automate refactoring code to use the new lowercased attributes.

adamchainz (commented):
Brilliant, thanks for getting this started as a DEP 🥳

draft/0015-content-types.rst
```diff
@@ -0,0 +1,149 @@
...
+- ``can_handle()`` should accept one argument being the ``media_type`` attempting to be parsed and returns a boolean to indicate if this parser can parse the given media type.
+- ``parse()`` accepts one argument being the ``data`` to parse and returns the parsed data.
+
+By default all requests shall be parsed by Django's default parsers. But to ensure backward compatability the new behaviour shall only be available when accessing the parsed response via the new ``data`` attribute.
+
+The new ``data`` attribute should parse requests in as close a way as possible to the current ``POST`` attribute. In addition it shall:
+
+* Parse ``application/JSON....`` types.
+* For ``multipart/form-data`` types each part will be parsed with the appropriate parser, if available.
```
orf, line 66: Do we need to specify anything about duplicate keys here? I.e what happens if `{foo: 1, foo: 2}` is sent via JSON or multipart, for JSON should we specify which `foo` value is chosen?

draft/0015-content-types.rst
```diff
@@ -0,0 +1,149 @@
...
+Abstract
+========
+
+Currently Django can parse requests for ``application/x-www-form-urlencoded`` and ``multipart/form-data`` types. Other types, such as JSON are currently returned as a string.
+
+This DEP proposes to add configurable content type parsers to allow parsing of additional content types. It is proposed that Django will include a parsing of JSON, and appriate hooks to allow users to add custom parsers for other content types.
+
+Parsed data from an ``HttpRequest`` is accessed via its ``POST`` attribute. It would be a breaking change if Django were to start parsing content types where currently a string is returned. To avoid introducing a breaking change it is proposed that a new ``data`` attribute is added for the new behaviour.
+
+While introducing a new name for ``POST`` it is proposed that the names for the other attributes are modernized with an equivalent behaviour.
```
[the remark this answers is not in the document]
carltongibson: I think `form_data` is confusing. 

In the first place, GET forms are a thing. But once you're parsing other content types from the body, there's no form even in play at all.

carltongibson:
Thanks for picking this up @smithdc1! 🎁 Good timing. Let's chat in Vigo.

draft/0015-content-types.rst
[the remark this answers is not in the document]
adamchainz: `post_form_data` ? I would just like another lowercase name for the existing attribute so it’s not left uppercase-only, requiring users to adopt `data`.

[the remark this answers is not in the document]
carltongibson: Oh, you mean **as well** as then adding the `data` attribute? I'd misunderstood. 

OK, yes, something like that makes sense. 

Let's discuss in Vigo, where we can likely bikeshed it to death over less than a single coffee ☕️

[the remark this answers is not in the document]
carltongibson: Having spoken with @adamchainz at DjangoCon, I agree with him. Adding `form_data` as an alias to `POST`, maintaining the existing behaviour is a good idea. 👍

draft/0015-content-types.rst
adamchainz: Can it just be "whatever json.loads does"?

orf: I guess: I was trying to think if there where any risks associated with this, perhaps that could lead to something similar to the [recent GitHub with bypass CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-7028), but I don’t think that’s the case.

The only way I can think that this would occur would be convoluted:
1. `{username: foo, username: bar}` is submitted
2. The last key is chosen by json.load, and `request.data[username]` returns “bar”
3. The body is forwarded or re-parsed, and that system choses the first key, resulting in username=foo

Since this can happen anyway, not sure it’s relevant here.

draft/0015-content-types.rst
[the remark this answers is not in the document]
smithdc1: In this case could this DEP become just about renaming? 

That would allow https://github.com/django/django/pull/17546 to make progress without this? 

What do you think?

draft/0015-content-types.rst
smithdc1: Initial work showing this is possible: https://github.com/smithdc1/django-upgrade/commit/c043761cb2fa00e97c3ea205be7e79fcec6f075d

draft/0015-content-types.rst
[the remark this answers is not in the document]
carltongibson: > What do you think?

My honest take is that we need both. I don't think renaming things without adding anything is great, and just sticking the `data` object on without updating the request seems an error. 

But, logically, yes, maybe. One strategy would be to plough on and see where we get.

collinanderson:
Hi All,

Not to sure where exactly to leave put this feedback but wanted to at least mention some concerns:

First of all I think keeping `form_data` as the exact same behavior as `POST` is a really good idea, along with not touching the almost 20-year old original names. I do think having two names for everything could be confusing, but the new names are probably more clear for beginners (who aren't PHP programmers like 20 years ago). However `meta` could probably use some improvement to be more clear: `cgi_variables`? `cgi_environ`? `cgi_environment_meta_variables`?

Regarding content-negotiation, I have some of concerns:

- If Django does any json parsing (including a `request.get_json()` attr), are there going be denial-of-service security issues? Like does Django need to respect DATA_UPLOAD_MAX_NUMBER_FIELDS? Does Django need to worry about how deeply nested the json can possibly go? Does Django need to worry about cases where field/key names are repeated? (I guess that one was already mentioned on Jun 7th.)

- json and form_data aren't really as interchangeable as much as urlencoded/multipart are. urlencoded and multipart both allow duplicate keys as a way of specifying a list (like multiple checkboxes or select multiple). It's all standardized as part of html, but there's no such thing as <form enctype="application/json">, so there's no way to auto-translate between the two. (Same with json and xml: they are not quite interchangeable.)

- I'm not sure it's generally a good idea to let the client decide what content-type to use when sending data. I could see some occasional cases to allow content negotiation, but I think in general it might be more secure-by-default to lock a view down to only accepting one content type? Like if my code processing form data from an html form is expecting only string-to-string key value pairs, it should probably stick to the .form_data attribute. Using a .data attribute and opening up the possibility of the client sending in json to that attribute might become a surprise issue. It might not even be dictionary.

- Similarly, people might set up Web-Application-Firewall rules to do some filtering on json that's getting passed to a view, and a client sending urlencoded or multipart might be able to bypass that filtering. (and vice-versa).

- Also re denial-of-service, I also think Django shouldn't parse json just because it was sent, unless the view is actually asking for the json. Like, any current view that's currently accepting form data shouldn't start auto-parsing json like it does with GET/POST/FILES/COOKIES/headers, just because Django starts to allow json. Wait until the view actually calls request.json() before parsing it. The view might not actually want it and parsing json is more complicated and slower than any of the other parsing, except maybe multipart, but there's already a bunch of denial-of-service checks for that.

- I kinda hinted at this before https://code.djangoproject.com/ticket/32646#comment:5, but it seems to me what people actually want is to have an easy way to parse json, but Django is trying to solve that by making a generic framework to handle "any" content type, which I personally think is actually going to lead to a lot of issues and confusion and wishy-washy views. If your view is expecting json then get it from `request.json()` (which should check content-type before parsing, like .form_data does). If you're expecting urlencoded/multipart, use `request.form_data`. If you're expecting something else, then parse it yourself. It makes the code more clear what's going on and therefore easier to read, and probably more secure.

- In general, if Django does allow this sort of auto-content negotiation, I'd personally recommend not using it by default, unless you know you actually need that feature and know what you're doing. This seems kinda similar to the issue with `request.REQUEST` (union of GET and POST) that was removed because of security issues.

- Explicit is better than implicit, Simple is better than complex, Readability counts, etc.

-  I don't think you need to add content-negotiation in order to justify renaming request vars, because I think the temptation will be for people to switch to using content-negotation as part of the rename because it's the shiny new exciting feature, and I don't think that's something that should be encouraged by default for the reasons above.

Anyway that's just a quick rant / two cents about content-negotiation, doesn't need to hold things up, but I thought I'd at least mention these issues for my own sake so they're not swirling around in my head forever.

Thanks,
Collin

smithdc1:
Thank you for your thoughts -- A couple of responses from me. I'm happy to add these to the DEP if that would be helpful. 

> If Django does any json parsing (including a request.get_json() attr), are there going be denial-of-service security issues? 

I think ensuring compliance with the [`DATA_UPLOAD_MAX_MEMORY_SIZE`](https://docs.djangoproject.com/en/5.0/ref/settings/#data-upload-max-memory-size) setting would be the defense here? This would be consistent with how other requests are currently handled. 

> Also re denial-of-service,  ....  Wait until the view actually calls request.json() before parsing it.

I agree that the behavior for `.data` should be the same as `.POST` and so on -- i.e. only parse when required.

draft/0015-content-types.rst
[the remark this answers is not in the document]
tim-schilling: @carltongibson @adamchainz any chance you remember that discussion and can recap it here? I share your original reservations about `form_data`. If someone learns Django while building an API, that's going to be confusing. Is the reason we're avoiding naming this something closer to the HTTP spec because we'll have `data` and we're trying to avoid confusion between those?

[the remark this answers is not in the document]
carltongibson: @tim-schilling I think this is the key point: 

> We can’t find-and-replace request.POST with request.data without adding new functionality unsafely.

The new `.data` attribute isn't behaviourally neutral. We need a safe migration path for existing code.

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+
+By default all requests shall be parsed by Django's default parsers. But to ensure backward compatability the new behaviour shall only be available when accessing the parsed response via the new ``data`` attribute.
+
+The new ``data`` attribute should parse requests in as close a way as possible to the current ``POST`` attribute. In addition it shall:
+
+* Parse ``application/JSON....`` types.
+* For ``multipart/form-data`` types each part will be parsed with the appropriate parser, if available.
+* Raise an ``UnsupportedMediaType`` (415) error if an unsupported content type is attempted to be parsed.
+
+The new data attribute is therefore not 100% equivalent to POST since it will parse JSON (and other data types as parsers for those are configured) where POST would return a string.
```
apollo13, line 55: Just reread the DEP due to a thread on the forum. Is this actually true? While it is true that the new `.data` attribute is not 100% equivalent it doesn't seem true that `POST` would ever return a string but rather an empty `QueryDict` if the content type is not something that it supports.

carltongibson: I think that's probably talking about JSON multiparts (but I'd need to look at tests to recall 100%)

LilyFirefly:
Something I think would be valuable here is discussion of HTTP verbs beyond just `GET` and `POST`, most notably `PUT` and `PATCH`. If I understand correctly, content negotiation is likely to most relevant for an API rather than a traditional web page. Along these lines, I think referencing prior art in Django Rest Framework and other third party API libraries would also improve the DEP.

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Custom parsers for additional content types shall be supported. To allow this it is proposed that ``HttpRequest`` adds a new property which returns a list of parsers to be used when attempting to parse a request.
+The list of parsers to be used can be set on the request, but must be done before ``data`` or ``FILES`` is accessed. This could be done in a middleware or in a view. For example::
+
+    def index(request):
+        request.parsers = [MyCustomParser(), FormParser(), ...]
+        ...
```
LilyFirefly, line 62: In https://github.com/django/django/pull/17546#issuecomment-1835758657 it's decided not to have a setting for this. I think we should preserve that decision and rationale in this DEP too.

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+* ``FILES`` -> ``files``
+
+The behaviour of renamed attributes shall be 100% compatible with the existing attributes. To ease migration to the new names django-upgrade can be used to automate refactoring code. Some work to show this is possible is available [1].
+
+[1] https://github.com/smithdc1/django-upgrade/commit/c043761cb2fa00e97c3ea205be7e79fcec6f075d
+
+Motivation
+==========
+
+This DEP is needed to add configurable content type parsing to Django. While this is likely to be well received the proposal also suggests adding new aliases which is more controversial.
```
LilyFirefly, line 81: I don't like that the DEP just assumes that configurable content type parsing is desirable. I think the reasons why we want this should be much more spelled out in this section.

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+   :depth: 3
+   :local:
+
+
+Abstract
+========
+
+Currently Django can parse requests for ``application/x-www-form-urlencoded`` and ``multipart/form-data`` types. Other types, such as JSON are currently returned as a string.
+
+This DEP proposes to add configurable content type parsers to allow parsing of additional content types. It is proposed that Django will include a parsing of JSON, and appriate hooks to allow users to add custom parsers for other content types.
```
tim-schilling, line 23:
```suggestion
This DEP proposes to add configurable content type parsers to allow parsing of additional content types. It is proposed that Django will include a parsing of JSON, and appropriate hooks to allow users to add custom parsers for other content types.
```

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+Motivation
+==========
+
+This DEP is needed to add configurable content type parsing to Django. While this is likely to be well received the proposal also suggests adding new aliases which is more controversial.
+
+The motivation to improve the names is that ``request.GET`` and ``request.POST`` are misleadingly named:
+
+* ``GET`` contains the URL parameters and is therefore available whatever the request method. This often confuses beginners and "returners" alike.
+
+* ``POST`` contains form data on ``POST`` requests, but not other kinds of data from ``POST`` requests. It can confuse users who are posting JSON or other formats.
```
tim-schilling, line 87: I don't disagree with this point, but I think we're trading sources of confusion. I think it's totally possible for a newer dev to misunderstand the differences between `request.form_data` and a `django.forms.Form`'s `data` and `cleaned_data`. We'd need them to know that `form_data` is the data you pass into the Form class. It's not your form's actual data to be used (`cleaned_data`)

```python
def view(request):
    if request.method == "POST":
        form = MyForm(request.form_data)
        if form.is_valid():
            some_value = form.cleaned_data["my_field"]
```

I'm not saying this needs to change. I want to see if anyone else feels the same or can assuage my concern.

draft/0015-content-types.rst
tim-schilling:
```suggestion
This DEP is needed to add configurable content type parsing to Django. Since Django takes a batteries included approach, supporting content type parsing for common content types such as ``application/x-www-form-urlencoded``, ``multipart/form-data`` and ``application/json`` is expected for a modern web application framework. It should be configurable so that the community can extend the content types that Django applications can support. This allows the eco-system the flexibility to grow with the rest of web development industry.

While this is likely to be well received the proposal also suggests adding new aliases which is more controversial.
```

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+
+The main objection received by the community is the renaming of the attributes. This causes a lot of churn in documentation to rename attributes where the behaviour of these is equivielent.
+
+Other options are:
+
+- Leave additional content type parsing to 3rd party packages, e.g. DRF
+- Introduce content type parsing and only add the new ``data`` attribute.
+
+The new names for unchanged attributes is proposed as it's considered this a worthwhile improvement in its own right and introduces consistent naming across ``HttpRequest`` attributes. That is, without renaming the change only the new ``data`` attribute would be an outlier.
+
```
tim-schilling, line 111:
```suggestion

A setting was considered for managing the content parsers for a Django application. The setting approach was dismissed because configuring the parsers is seen as something rare. It would be similar to the `FILE_UPLOAD_HANDLERS` setting which is rarely changed from the default. The solution is to have the application configure the parsers in the view or early in the middleware.
```

@LilyFoote Regarding your comment here: https://github.com/django/deps/pull/88/files#r2041206276

draft/0015-content-types.rst
tim-schilling: Ack, ignore this. There's a different thread.

draft/0015-content-types.rst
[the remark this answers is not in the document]
tim-schilling: Thank you for that Carlton. That makes total sense.

However, I think we're trading sources of confusion by adding `form_data`. I think it's totally possible for a newer dev to misunderstand the differences between `request.form_data` and a `django.forms.Form`'s `data` and `cleaned_data`. We'd need them to know that `form_data` is the data you pass into the Form class. It's not your form's actual data to be used (`cleaned_data`)

```python
def view(request):
    if request.method == "POST":
        form = MyForm(request.form_data)
        if form.is_valid():
            some_value = form.cleaned_data["my_field"]
```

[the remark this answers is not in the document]
collinanderson: > ...naming this something closer to the HTTP spec...

The content-type is either `application/x-www-form-urlencoded` or `multipart/form-data` so `request.form_data` is pretty close, right?  If we try to be sure to always refer to it as `request.form_data`, not just `form_data`, that might help. A full name could be "request body form data".

Edit: `request.post_form_data` is also probably fine and more clear, because in html forms you can't use method="PUT" or anything else.  (or `request.body_form_data`??)

> ...differences between `request.form_data` and a `django.forms.Form`'s `data`...

- In normal cases `request.form_data` and `form.data` will be the same thing, right?

(I still personally think that `application/json` should end up as `request.json`, like how `requests`' `response.json()` works. Same with `response.json()` in javascript's `fetch()`. I think it would be more clear than an intentionally generic `request.data` which could be confusing.)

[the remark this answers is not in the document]
merwok: `request.json` is also found in Pyramid, alongside `request.body` (bytes) and others.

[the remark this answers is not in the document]
tim-schilling: > In normal cases request.form_data and form.data will be the same thing, right?

Yes that's true. I suppose the fallout of this confusion is smaller than I originally thought. Good call out.

draft/0015-content-types.rst
LilyFirefly: This is good, thanks!

I think the wording can be improved in one place - it talks about configuring parsers being rare, but then compares to `FILE_UPLOAD_HANDLERS` being frequently changed from the default. I think I get what was intended here, but rewording would make it easier to understand.

tim-schilling: Whoops! Good catch. Fixed in the suggestion to say "setting which is rarely changed"

LilyFirefly: Works for me now! :rocket:

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+Parsed data from an ``HttpRequest`` is currently accessed via its ``POST`` attribute. It would be a breaking change if Django were to start parsing content types where currently a string is returned. To avoid introducing a breaking change it is proposed that a new ``data`` attribute is added for the new behaviour.
+
+While introducing a new name for ``POST`` it is proposed that the names for the other attributes are modernized with an equivalent behaviour:
+
+* ``GET`` -> ``query_params``
+* ``POST`` -> ``form_data``
+* ``COOKIES`` -> ``cookies``
+* ``META`` -> ``meta``
+* ``FILES`` -> ``files``
```
nanuxbe, line 33: OH: One of these is not like the others.

While those are all attributes of the current request, of those are related to the content of the request while `META`/`meta` is not and is mostly provides things not really "submitted", with the notable exception of `headers`.

Would leaving `META` alone and extracting `headers` be something that would make more sense?

TeaDrinkingProgrammer:
I would just like to add a user perspective that I haven't seen highlighted yet. In discussions around content type parsing, it is often that forms always use `POST` since you cannot use `PUT` or `PATCH` in an HTML form. [This was mentioned](https://groups.google.com/g/django-developers/c/dxI4qVzrBY4/m/m_9IiNk_p7UJ) as the reason why PUT requests do not populate the POST dictionary. While this is the case with standard HTML5 forms, hypermedia libraries like HTMX encourage users to use the most idiomatically appropriate HTTP verb for a request, which sometimes means using `PUT` or `PATCH`. Of course, you can still use a `POST` request, but I would like to be able to use a `PUT` when I am updating an existing resource, for example.

LilyFirefly:
This DEP should be re-numbered (I suggest 0017) because we've accepted #98 at DEP 0015.

camuthig:
Hey all, this DEP has been presented during forum discussions about better supporting API development with Django as [a good first step](https://forum.djangoproject.com/t/django-needs-a-rest-story/42814/55).

If there is a good way for me to help, I would like to.

Does this DEP have a shepherd yet?
@smithdc1 would like like any help authoring the DEP or moving anything along?

smithdc1:
Hi @camuthig 

Help with pushing this forward would be very much appreciated.  

Do you have thoughts on what steps should be taken to progress this?

camuthig:
@smithdc1 if there isn't a shepherd, I think it would be best to find one. I am not capable of filling that role, though.

@carltongibson sent me to this PR, and I know he has been a part of this discussion for a very long time. I'm not sure if he is available to be a shepherd or if he might be able to recommend someone else in the community for the task. 

I think first steps are to review the suggestions made already in the PR and either accept them or work through them. After that, I think we would need to see where things stand, asking the reviewers to brush up on the conversation and see if conceptually the DEP can move forward or if there are deeper discussions that need to be had on top of the recommendations made here.

I don't have the ability to accept suggestions. I can review suggestions and provide a 👍 or ask questions as a marker for you? From my review so far most seem like good recommendations supported by multiple reviewers, though, so it is likely they are ready for acceptance already?

carltongibson:
I don't have capacity to lead this now. 

I have a plan to implement what's needed as a third-party handler subclass and request class, to act as a proof of concept and test bed. I'd then swing back to a DEP once it was demonstrated in practice. 

I would have got to that already, except it hasn't managed to become a priority for me. It's not something I've actually needed on the work project (and we're nearly three years in at this point). We will eventually, but I can say when that will be. 

I believe @nanuxbe is interested in seeing this progress, but I also know she has several projects in hand. 🤹

> ... implement what's needed as a third-party handler subclass and request class, to act as a proof of concept and test bed.

If you want to push it forwards, I'd advise doing that. The existing PRs already show the way, there's everything you need there. (It'd be a good learning experience too.)

camuthig:
That's a great recommendation @carltongibson . Thank you for providing the feedback.

@smithdc1 are you open to working on a third-party implementation? I am traveling for a few days but I can start working on something later this week as an experiment with the ideas you have described here.

ehmatthes:
> This DEP should be re-numbered (I suggest 0017) because we've accepted https://github.com/django/deps/pull/98 at DEP 0015.

Yes, I had a hard time figuring out where to make comments about the new project template because I kept ending up here.

carltongibson:
I changed the PR title at least.

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
+DEP 0015: Content type aware parsing and modernization of the HttpRequest API
```
LilyFirefly, line 1:
```suggestion
DEP 0017: Content type aware parsing and modernization of the HttpRequest API
```

draft/0015-content-types.rst
```diff
@@ -0,0 +1,131 @@
...
+=============================================================================
+
+:DEP: 0015
```
LilyFirefly, line 4:
```suggestion
:DEP: 0017
```

LilyFirefly (changes requested):

camuthig:
I have started a proof-of-concept implementation as a package for this. It isn't ready for prime time yet, but I wanted to share it here for visibility.

https://github.com/camuthig/django-content-type-parsing

I just noticed the changes happening in https://github.com/django/django/pull/20498 . I still need to review that feature request and pull request to understand how it plays along with this DEP.

jonbiemond:
Regarding request data attribute naming, can we align the new HTTP QUERY method implementation with this proposal?

As part of implementing the QUERY method, there is a question as to which attribute of `HttpRequest` Django should populate with the QUERY request content. For more detail, see the discussion on the ticket: https://code.djangoproject.com/ticket/37232

It would be a win if the QUERY implementation creates little to no extra work for this DEP. What attribute name would this DEP propose for the new QUERY method? Would it use `query_params` like GET?

Thanks to nanorepublica for the suggestion.

carltongibson:
Hey @jonbiemond... so QUERY allows populating the request body on what would otherwise be GET requests, which don't allow that, right? 

So `request.body`, but then whatever parsed equivalents of that we end up adding here. (Which isn't 100% decided.)

I think prior to any changes here, saying parse the body yourself might not be a bad answer. (QueryDict(...) isn't that much to ask.) 

I would think request.POST wouldn't be a good place for it. 

(I suspect it's a bit more complex than that. Do we need to pick this DEP up?)

jonbiemond:
Hi @carltongibson, that's right. QUERY is GET with content, meant to be safe and idempotent.

I think `request.body` is fine, and allows devs to start using the new method in Django. And it shouldn't add any extra migration effort to this DEP.

But I'm wondering if we can go even a step further, in an effort to nudge this DEP forward, by adding a new lowercase attribute name that aligns with this proposal. I do think the implication then is that Django automatically parses QUERY content into that new attribute, like we already do for POST. To ask an explicit question, is that too big of a leap to make at this stage of the new QUERY spec as well as for this DEP?

> I think prior to any changes here, saying parse the body yourself might not be a bad answer. (QueryDict(...) isn't that much to ask.)

I figure this probably answers the question.

carltongibson:
I think doing it by half leaves the API looking weird. Why just one lowercase attribute? 

We should modernise the request object api, absolutely. But not by just adding one attribute.

jonbiemond:
I think that's reasonable. Thanks! I will voice my vote to have the QUERY method populate `request.body`. And I hope this DEP will one day give me a way to conveniently access the parsed data.

PR #90, opened 2024-06-01
Tidy DEP 14 a bit

adamchainz:
1. Indent Python code with four spaces.
2. Modernize and complete the type hints, specifically replacing `Dict` and `List` with `dict[k, v]` and `list` (or `tuple[Any, ...]` in the case of `Task.args`).
3. Tidy language around asynchronous tasks, including expanding the “async” shorthand and removing its syntax highlighting.
4. Replace the term “globally-importable” with “module-level”, which is what I understood was meant.


adamchainz:
@RealOrangeOne

RealOrangeOne (approved):
I have absolutely no idea how the indenting got this messed up, but thanks for cleaning it up!

I also the changed wording is much clearer - thanks again.

RealOrangeOne:
As you're already here, it'd be great if you could amend https://github.com/django/deps/pull/86#discussion_r1623213992 at the same time! :slightly_smiling_face:

adamchainz:
> As you're already here, it'd be great if you could amend [#86 (comment)](https://github.com/django/deps/pull/86#discussion_r1623213992) at the same time! 🙂

Done, merging!

PR #93, opened 2024-09-20
Made cosmetic edits to the technical board role description in DEP 10.

sarahboyce:
The Django docs uses this sentence in the Steering Council [role description](https://docs.djangoproject.com/en/5.1/internals/organization/#id5)

I wanted to tweak the role description because saying `takes part in filling certain roles` sounds vague within a role description (even though it reads perfectly fine here). 

See: https://github.com/django/django/pull/18607

I want to keep this DEP and the Django docs role description aligned. So, I would like us to tweak the wording here (so I can also tweak the wording there).

I imagine "takes part in filling certain roles" is a catch all which refers to things like adding or removing releasers.
Another option is removing `takes part in filling certain roles` part.

### DIFF
--- a/final/0010-new-governance.rst
+++ b/final/0010-new-governance.rst
@@ -351,8 +351,8 @@ Repurposed role: Technical Board
 
 The Technical Board provides oversight of Django's development and
 release process, assists in setting the direction of feature
-development and releases, takes part in filling certain roles, and has
-a tie-breaking vote when other decision-making processes fail.
+development and releases, selects Mergers and Releasers, and has a
+tie-breaking vote when other decision-making processes fail.
 
 The powers of the Technical Board are:
 


LilyFoote (approved):

ubernostrum:
The "filling certain roles" bit is about the Mergers and Releasers, maybe edit it to make that clear?

sarahboyce:
> The "filling certain roles" bit is about the Mergers and Releasers (since they have to be confirmed by a vote of the technical board under DEP 10), maybe edit it to make that clear?

Ah thank you - updated :+1: 
Let me know if you're happy

Previously I read it like "The Technical Board fills certain roles (themselves)", but now I see I read it wrong and it should be read like "The Technical Board fills certain roles (with other contributors)"

LilyFoote (approved):

ubernostrum (approved):

PR #91, opened 2024-09-10
Update DEP 7 terminology to match current landscape

thibaudcolas:
This is a follow-up to @jacobian’s #81, updating DEP 7 so its terminology matches current operations. 

As of today I believe the relevant projects are:

- https://github.com/django/django-localflavor
- https://github.com/django/channels
- https://github.com/django/daphne/
- https://github.com/django/asgiref/
- https://github.com/django/channels_redis/

This policy also came up as part of Django ticket [#33620 Accessibility in pipeline](https://code.djangoproject.com/ticket/33620).

PR #95, opened 2024-10-16
Update DEPs terminology to match current landscape

thibaudcolas:
Follow-up to #81, and #91, updating the terminology in use to reflect current governance, with the exception of DEP 10.

### DIFF
--- a/draft/0007-dependency-policy.rst
+++ b/draft/0007-dependency-policy.rst
@@ -25,8 +25,8 @@ policy of "no external dependencies allowed."
 In a nutshell, the policy is that adding a new external dependency should be
 treated similarly to adding a major new feature to Django: it requires a DEP,
 demonstration that the dependency is needed, rough consensus among the
-community and core team that the chosen dependency, and a final decision by
-the Technical Board.
+community and core contributors that the chosen dependency,
+and a final decision by the Steering Council.
 
 The rest of this document explains the guidelines and process for adding new
 dependencies, as well as the background and motivation about why this policy was
@@ -56,7 +56,7 @@ as Django itself. We define "maturity" as:
   issues that wouldn't make it into Django shouldn't be accepted as a dependency,
   either.
 
-- **Maintained** - if we discover bugs in a dependency, we need to be fairly 
+- **Maintained** - if we discover bugs in a dependency, we need to be fairly
   confident that they'll be fixed quickly.
 
 - **Takes security seriously** - we should be confident that if we or our users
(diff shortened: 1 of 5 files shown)


tim-schilling (approved):
Thank you for updating these references Thibaud.

nanuxbe:
+1

carltongibson (approved):
Yep great thanks. 👍

draft/0007-dependency-policy.rst
```diff
@@ -123,11 +123,11 @@ Re-evaluating dependencies
 --------------------------
 
 During each minor release cycle -- and especially before LTS releases -- the
-core team should re-evaluate all existing dependencies. If some dependency is
-starting regress on the maturity front (particularly if it has become
-unmaintained), we want to identify it early and start looking for backup plans.
-This might mean removing the dependency, taking over maintenance ourselves, 
-looking for funding to pay new maintainers, etc.
+core contributors should re-evaluate all existing dependencies. If some
```
LilyFoote, line 126: I wonder if this should be more specific:

```suggestion
Steering Council and Fellows should re-evaluate all existing dependencies. If some
```

Otherwise, it's not really clear whose responsibility it is.

But maybe this is straying a bit far from the point of this particular PR...

draft/0007-dependency-policy.rst
```diff
@@ -151,8 +151,8 @@ occasional downtime, leading to frustration when trying to deploy to production.
 didn't go well. Many users struggled even to get the package installed.
 
 That experience led to a deep suspicion of Python packaging tools among the
-Django core team, and a de-facto policy emerged of only requiring the lowest
-common denominator: installation via direct download and ``python setup.py
+Django core contributors, and a de-facto policy emerged of only requiring the
+lowest common denominator: installation via direct download and ``python setup.py
```
LilyFoote, line 155: Given this is a historical section I think it would be OK not to update the terminology here. That said, that's not a strong opinion.

draft/0007-dependency-policy.rst
LilyFoote, line 1: I notice we have two DEP 7s:

* `final/0007-official-projects.rst`
* `draft/0007-dependency-policy.rst`

It's probably out of scope of this PR to fix this, but I think the draft DEP 7 should be re-numbered.

LilyFirefly (approved):
Line 74 of the accessibility team DEP still references the Technical Board:

> Technical Board as required. To begin, several areas have been identified:

draft/0007-dependency-policy.rst
thibaudcolas: Yes, this seems like a reasonable change to me but too substantive for what I was going for here. If we didn’t think the change I went for is appropriate, I’d rather remove all changes to this DEP.

tim-schilling:
Thank you @thibaudcolas! I went ahead and merged it to remove that bit of decision making from your plate.

PR #100, opened 2025-01-26
Add django command

ryanhiebert:
Motivated by a desire to remove confusing papercuts in Django and to follow common convention in the Python ecosystem, this DEP proposes to add a new ``django`` command equivalent to the existing ``django-admin`` command, and to update the documentation to prefer this new spelling.

[Forum discussion](https://forum.djangoproject.com/t/name-the-main-command-django/37230)
[Rendered](https://github.com/ryanhiebert/deps/blob/django-cmd/draft/0000-name-main-command-django.rst)


draft/0000-name-main-command-django.rst
```diff
@@ -0,0 +1,212 @@
...
+and the benefits of following common convention again outweight the cost.
+
+Ambiguity
+---------
+
+The ``django`` command can be seen as a terminology conflict
+with the name of the Python package or the name of the project itself.
+However, Django is an exception to what most tools with a CLI do,
+so the current situation is already confusing to new users.
+It is worth the trade some confusion over this ambiguity
```
shaib:
```suggestion
It is worth to trade some confusion over this ambiguity
```

shaib (commented):
We can mitigate some of the "multiple spelling" issue, by having `django-admin` not as a completely equivalent alias to `django`, but as a thin wrapper that invokes `django` and prints a message along the lines of

>  The `django-admin` command is being renamed to `django`. You can keep using either name, they are equivalent except for the printing of this message. For more details on the naming change, see (link to DEP 100).

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,214 @@
...
+this DEP proposes to add a new ``django`` command equivalent to
+the existing ``django-admin`` command,
+and to update the documentation to prefer this new spelling.
+
+Specification
+=============
+
+The ``django`` command will be added as the preferred spelling
+for the existing ``django-admin`` command.
+The ``django-admin`` command will remain indefinitely,
```
RealOrangeOne, line 32: > Indefinitely

That's a long time! Should there be consideration for a deprecation window, even if it's longer than a standard one might be.

ryanhiebert:
> a thin wrapper that invokes `django` and prints a message

Thanks for that suggestion. I've worked it in. I think it'll make it a little easier to start any discussion about it.

draft/0000-name-main-command-django.rst
```diff
@@ -0,0 +1,214 @@
...
+
+The ``django`` command will be added as the preferred spelling
+for the existing ``django-admin`` command.
+The ``django-admin`` command will remain indefinitely,
+with a message that says
+
+> The ``django-admin`` command is being renamed to ``django``.
+> You can keep using either name,
+> they are equivalent except for the printing of this message.
+> For more details on the naming change, see DEP XXXX.
```
LilyFirefly: This isn't rendering correctly in the rich diff.

draft/0000-name-main-command-django.rst
```diff
@@ -0,0 +1,214 @@
...
+
+Official documentation will be updated
+to reference this new ``django`` command
+everywhere that ``django-admin`` is currently referenced
+by searching the repository for ``django-admin``
+and auditing each instance to determine
+if it is a reference to the command and should be updated.
```
RealOrangeOne: Nit: It's probably not necessary to define _intelligent search and replace_.
```suggestion
everywhere that ``django-admin`` is currently referenced.
```

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,214 @@
...
+that the two commands are equivalent,
+the added messaging in the ``django-admin`` command,
+and because the benefits of
+following common convention outweigh the cost.
+
+Ambiguity
+---------
+
+The ``django`` command can be seen as a terminology conflict
+with the name of the Python package or the name of the project itself.
```
RealOrangeOne, line 157: I think it might be worth highlighting here that the conflict is actually a feature.

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,214 @@
...
+However, Django is an exception to what most tools with a CLI do,
+so the current situation is already confusing to new users.
+It is worth the trade of some confusion over this ambiguity
+for the clarity gained by consistency with other tools.
+Over time, usage of the current command will be less common,
+and the confusion will be less likely to surface.
```
RealOrangeOne, line 165: Suggestion: It might be interesting to call out the existing ambiguity between `django-admin` and the Django Admin interface. "Using the Django admin" can easily mean the web interface or the CLI, and new users could easily choose the wrong one.

RealOrangeOne (commented):
Love this idea :heart:

accepted/0016-name-main-command-django.rst
jefftriplett: I suspect to >99% of the Django population, saying "Django Admin" defaults to the free CRUD view we give people for managing their website. Vs. teaching them it's how you start a Django website, and then you'll never use it again.

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,214 @@
...
+and some broader examples include ``ember``, ``rails``, and ``vite``.
+
+The broad acceptance of this pattern has been reinforced
+by tools like ``uv tool run`` (aka ``uvx``) and ``pipx``,
+where commands that are the same as the package name
+get special privileges and a simpler syntax.
+For example ``uvx pytest`` automatically downloads and runs
+the ``pytest`` command from the ``pytest`` package.
+
+This pattern has been further reinforced
```
jefftriplett, line 81: I'm not sure where it belongs, but it's a common tripping point for folks on the forum who see `No module named django-admin` because they try to run `python -m django-admin` because we teach people to start with the `django-admin` command. This is because Python has a best practice of using `python -m pip` (which you are talking about next) to avoid accidentally installing a library like Django globally. 

I have seen people fiddling with `uv tool run --from django django-admin` to bootstrap a project because it wasn't obvious that `uv tool run django` would work instead because we train everyone to start with `django-admin`.

draft/0000-name-main-command-django.rst
```diff
@@ -0,0 +1,214 @@
...
+and some broader examples include ``ember``, ``rails``, and ``vite``.
+
+The broad acceptance of this pattern has been reinforced
+by tools like ``uv tool run`` (aka ``uvx``) and ``pipx``,
+where commands that are the same as the package name
+get special privileges and a simpler syntax.
+For example ``uvx pytest`` automatically downloads and runs
+the ``pytest`` command from the ``pytest`` package.
+
+This pattern has been further reinforced
```
ryanhiebert: I ended up including this in the "alternatives" section, discussing why just adding the alias didn't get us everything we want. I can move it to the main motivation section instead.

accepted/0016-name-main-command-django.rst
ryanhiebert: So far the feeling has been that the maintenance burden isn't significant enough to warrant removing such a widely documented feature in tutorials across the web. I'm inclined to keep it around forever.

draft/0000-name-main-command-django.rst
ryanhiebert: Thanks, fixed it!

accepted/0016-name-main-command-django.rst
ryanhiebert: I've added a bit more to the motivation section about this.

accepted/0016-name-main-command-django.rst
ryanhiebert: Added some words about this ambiguity.

accepted/0016-name-main-command-django.rst
ryanhiebert: I've added a section motivating this choice.

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,230 @@
...
+The correspondance between the command name and the package name
+allows for a more intuitive mapping to the alternative style.
+
+It is a common tripping point for folks on the forum to see
+``No module named django-admin``
+because they try to run ``python -m django-admin``.
+Others have had difficulty identifying
+the correct ``uvx`` (``uv tool run``) command as
+``uvx --from django django-admin``
+instead of ``uvx django-admin`` or ``uv django``.
```
merwok, line 96: Did you mean `uvx django` ?

ryanhiebert: Yes!

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,230 @@
...
+with messaging about the new name.
+There are no plans to remove the ``django-admin`` alias,
+because of the expansive amount of external resources,
+such as blog posts and tutorials,
+that reference the existing command name.
+
+Motivation
+==========
+
+Django is how many people first learn Python,
```
ulgens, line 60: Is this completely anecdotal, or do we have any data points from a survey?

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,230 @@
...
+``python -m pip`` to ensure that
+you're using the version of the module
+that is associated with your intended Python interpreter,
+and to avoid accidentally installing a library like Django globally.
+The correspondance between the command name and the package name
+allows for a more intuitive mapping to the alternative style.
+
+It is a common tripping point for folks on the forum to see
+``No module named django-admin``
+because they try to run ``python -m django-admin``.
```
tim-schilling, line 92: There are a total of 13 questions asked, for 3600+ views starting back in 2016. Typically, I would say this is small, but if we look at it from the perspective of first-time django devs, it seems to be common enough for us to improve.

- https://stackoverflow.com/search?q=%22No+module+named+django-admin%22
- https://forum.djangoproject.com/search?q=%22No%20module%20named%20django-admin%22
- https://groups.google.com/g/django-users/search?q=%22No%20module%20named%20django-admin%22
- Django Discord shows 6 questions with `"No module named django-admin"`
- Python Discord shows 1 question with `"No module named django-admin"`

accepted/0016-name-main-command-django.rst
jefftriplett: We could add it to future surveys if more people think it'd be a useful data point. 

My data point is asking while on the stage at DjangoCon US and seeing over half the room hold up their hands every time I ask them if they learned Python by learning Django.

accepted/0016-name-main-command-django.rst
```diff
@@ -0,0 +1,230 @@
...
+Official documentation will be updated
+to reference this new ``django`` command
+everywhere that ``django-admin`` is currently referenced.
+The implementor will coordinate with the translation team
+to assist in making all necessary translation updates.
+
+Backwards Compatibility
+=======================
+
+The existing ``django-admin`` command will remain indefinitely
```
tim-schilling, line 49: Personally, I'd like to see this not be the case. Having the message when people use `django-admin`, but never planning on removing it feels like unnecessary noise. [Or have the warning message removed until we have a plan in place. Otherwise we're going to end up with another setting to disable that messaging.]

ryanhiebert: There are competing community values that I'm trying to navigate:

1. Introduce the preferable spelling, `django`
2. Avoid breaking existing material that uses `django-admin`
3. Have only one way to do it
4. Avoid noisy warnings that don't have a clear time horizon

I don't see a way to have all four of these at the same time. If we skip the warning, we have more than one way to do it. If we set a clear duration on the warning then we will break existing material at that point in the future.

The persistent warning without a time horizon felt like a reasonable way to soft-enforce "one way to do it" without breaking existing material; adding a setting could further soften the enforcement. The warning is visible to developers using Django, without breaking any commands. That can give us some time to have a better sense of whether the community will ever want to remove the old way, or whether we'd rather compromise on "only one way to do it".

If the project is prepared to make that choice now rather than wait, I'm happy for that. I think that my preference would be the same as yours: let's plan to remove it at some specific point in the future. It might deserve a longer deprecation period due to the centrality of the command, although I think I'd personally be inclined stick with the policy as stated.

I suspect this DEP won't be able to be resolved without the SC determining the right way to balance these competing values, because key members of the community disagree on which values should have the highest priority for this question. Would it be best for the purpose of lining this question up for the SC to modify this DEP with a more aggressive deprecation suggestion, or keep it as-is?

LilyFirefly: I'm happy with either the current proposal or with setting an explicit timeline, though I think it should be longer than our usual deprecation cycles to allow the tutorial ecosystem to migrate.

jefftriplett: +1 to phasing `django-admin` out over a really long time. I agree with Tim that we should have an appropriate warning message, but my gut feel is that we should phase it in a bit slower than we might otherwise.

nanuxbe (approved):
As others have mentioned, my only concern with this is the lack of planned deprecation for the django-admin command.
I agree that this will need non-standard, longer term deprecation than the usual, so it might be best to come up with that in a separate PR/discussion/place

LilyFirefly (approved):

frankwiles (approved):

pauloxnet (approved):

accepted/0016-name-main-command-django.rst
adamchainz: I'm +1 for keeping `django-admin` indefinitely. pytest still has its legacy `py.test` command with no plans to deprecate, and it doesn't seem to harm users. What does hurt is when production scripts with no version control break due to the command going missing. In such cases, deprecation warnings are often never seen because they're piped into ignored logs.

But I'd also be okay with something quite long-term, like "After two major versions, it will be deprecated" would be okay. We can use language similar to [Python 3.14's announcement](https://docs.python.org/3/whatsnew/3.14.html#pep-784-zstandard-support-in-the-standard-library:~:text=Any%20deprecation%20or%20removal%20of%20the%20existing%20compression%20modules%20will%20occur%20no%20sooner%20than%20five%20years%20after%20the%20release%20of%203%2E14) of the new `compression` module:

> Any deprecation or removal of the existing compression modules will occur no sooner than five years after the release of 3.14.

draft/0000-name-main-command-django.rst
```diff
@@ -0,0 +1,230 @@
+========================================
+DEP XXXX: Name the main command `django`
+========================================
+
+:DEP: XXXX
+:Author: Ryan Hiebert
+:Implementation Team: Ryan Hiebert
+:Shepherd: Tom Carrick
+:Status: Draft
```
LilyFirefly:
```suggestion
:Status: Accepted
```

accepted/0016-name-main-command-django.rst
LilyFirefly, line 1: Move this to `accepted/0016-name-main-command-django.rst`.

draft/0000-name-main-command-django.rst
```diff
@@ -0,0 +1,230 @@
+========================================
+DEP XXXX: Name the main command `django`
```
LilyFirefly:
```suggestion
DEP 16: Name the main command `django`
```

draft/0000-name-main-command-django.rst
```diff
@@ -0,0 +1,230 @@
...
+========================================
+
+:DEP: XXXX
```
LilyFirefly:
```suggestion
:DEP: 16
```

LilyFirefly (changes requested):
This DEP is now accepted. All members of the Steering Council are in favour (I've spoken with @carltongibson in person).

@ryanhiebert @knyghty Can one of you make the requested changes to this PR? Then it can be merged.

LilyFirefly:
@tim-schilling Ooops, I thought you had approved this, but I just noticed you haven't actually. Are you happy to approve?

tim-schilling (approved):
I technically approved in Slack. All good here.

knyghty:
Amazing, thank you everyone!

Antoliny0919:
Amazing, I’m looking forward to this change 🙌🙌

accepted/0016-name-main-command-django.rst
ulgens: Thank you. I feel like this mention is worth having a reference for, but I guess it's too late now.

PR #103, opened 2025-11-06
Added a draft introduction for the Hybrid Signal handler

Arfey:
Introduce a draft proposal for the Hybrid Signal handler, which allows defining separate sync and async implementations for signal receivers to improve performance and simplify mixed context execution.

### DIFF
--- a/draft/xxxx-hybrid-signal.md
+++ b/draft/xxxx-hybrid-signal.md
@@ -1,0 +1,124 @@
+---
+DEP: XXXX Hybrid Signals
+Author: Mykhailo Havelia
+Implementation Team:
+Shepherd:
+Status: Draft
+Type: Feature
+Created: 2025-11-06
+Last-Modified: 2025-11-06
+---
+
+# DEP XXXX: Hybrid Signal
+
+Table of Contents
+- [Abstract](#abstract)
+- [Motivation](#motivation)
+- [Specification](#specification)
+- [Rationale](#rationale)
+- [Backwards Compatibility](#backwards-compatibility)
+- [Copyright](#copyright)
+
+## Abstract
+
+This DEP proposes the introduction of hybrid signal handlers in Django - signal
+receivers that can define different implementations for synchronous (`send`) and
+asynchronous (`asend`) dispatch. The goal is to eliminate the need for expensive
+`sync_to_async` and `async_to_sync` conversions when signals are triggered in mixed
+contexts, while keeping backward compatibility with existing Signal usage.
+
+## Motivation
+
+Currently, Django's signal system doesn't distinguish between synchronous and
+asynchronous receivers. A signal can connect either sync or async functions, but
+internally Django uses thread-based wrappers (`sync_to_async` or `async_to_sync`) when
+dispatching to handlers that don’t match the context type.
+
+Example:
+
+```python
+from django.core.signals import request_finished
+from django.dispatch import receiver
+
+@receiver(request_finished)
+def request_logger(sender, **kwargs):
+    print("Request finished")
+```
+
+When this signal is triggered using `asend`, Django wraps sync functions using
+`sync_to_async`, spawning a thread for each invocation. This adds unnecessary overhead.
+A hybrid signal would allow developers and third-party libraries to register both sync
+and async versions of a handler explicitly. This way, Django can directly dispatch to
+the appropriate function without any cross-context wrapping or thread spawning.
+
+# Specification
+
+A hybrid signal handler can define both sync and async implementations under a single
+registration.
+
+Example API:
+
+```python
+class HybridSignalHandler:
+
+    def sync_call(self, receiver, **kwargs):
+        raise NotImplementedError(
+            "send is not supported for this handler"
+        )
+
+    async def async_call(self, receiver, **kwargs):
+        raise NotImplementedError(
+            "asend is not supported for this handler"
+        )
+```
+
+Usage:
+
+```python
+@receiver(request_finished)
+class RequestLogger(HybridSignalHandler):
+    def sync_call(self, sender, **kwargs):
+        print("Request finished (sync)")
+
+    async def async_call(self, sender, **kwargs):
+        print("Request finished (async)")
+```
+
+## Rationale
+
+Django’s current signal system calls receivers the same way in all contexts.
+When an async signal is sent, Django wraps sync receivers with `sync_to_async()`,
+which runs them in a thread. This works but adds unnecessary overhead for short signals
+and makes async code less efficient. The idea of hybrid signals is simple: allow
+developers to define both sync and async versions of the same handler. Then Django
+can call the right one directly, without using wrappers or threads.
+
+This design:
+
+- avoids extra thread spawning and context switching
+- keeps signals simple and backward compatible
+- and gives developers clear control over how their handlers run
+
+Similar patterns already exist in other frameworks, for example in `Strawberry GraphQL`,
+where extensions can define both `resolve()` and `resolve_async()`.
+
+
+## Backwards Compatibility
+
+This proposal is fully backward compatible:
+
+- Existing Signal and receiver decorators continue to work as is.
+- Libraries can safely migrate incrementally.
+
+Potential risks:
+
+- None for existing code.
+- Minimal risk if developers misdefine hybrid handlers (guarded by `NotImplementedError`).
+
+## Copyright
+
+This document has been placed in the public domain per the Creative
+Commons CC0 1.0 Universal license
+(<http://creativecommons.org/publicdomain/zero/1.0/deed>).
+
+(All DEPs must include this exact copyright statement.)


draft/xxxx-hybrid-signal.md
```diff
@@ -0,0 +1,124 @@
...
+
+    def sync_call(self, receiver, **kwargs):
+        raise NotImplementedError(
+            "send is not supported for this handler"
+        )
+
+    async def async_call(self, receiver, **kwargs):
+        raise NotImplementedError(
+            "asend is not supported for this handler"
+        )
```
tim-schilling, line 72: Generally, this DEP is a good idea IMO :rocket:  I'm a little stuck on this implementation and wanted to see if you had thought of other approaches.

Did you consider revising the `@receiver` decorator to control whether the handler should be called based on the context? For example:
```python
@receiver(signal, sync=False, _async=True)
async def async_call(receiver, **kwargs):
    ...

@receiver(signal, sync=True, _async=False)
def sync_call(receiver, **kwargs):
    ...
```

This would avoid requiring people to use a inheritance in their handlers and still provide the flexibility you're looking for. I haven't thought this entirely through though, so I could be missing something.

Arfey: I considered that approach, but it's much easier to make mistakes with it. In most cases, a handler should run regardless of whether the context is async or sync. For example:

```python
@receiver(request_finished, with_async=False)
def close_db_connection(sender, **kwargs):
    # close sync DB connection
    ...
```

When `ASGIHandler` calls `signals.request_finished.asend(sender=self.__class__)`, Django will not close a database connection that was created inside `sync_to_async`.

```python
async def index(request):
    # do async work
    ...
    await sync_to_async(do_sync_work_with_sync_db)()
```

Additionally, even if production uses only async connections, the development server wraps all async views with `async_to_sync` and still sends request_finished synchronously.

That’s why I think using a class-based design is a better choice. It encourages developers to implement both async and sync versions explicitly, and Django can easily raise an exception if one is missing.

We can skip inheritance and just use duck typing (if an object has `sync_call` and `async_call`, then it is a hybrid handler) 😁

I don’t know, I just kind of like the idea of using classes — it feels like a better fit 😅
 
```python
class LogginHandler:

   def log(self):
       print('done')

    def sync_call(self, receiver, **kwargs):
        self.log()

    async def async_call(self, receiver, **kwargs):
        self.log()
```

carltongibson: Here it seems like we want the accept a pair of callables, for the sync and async case respectively, rather than just the one. 

The old fashioned way to do that might have been to pass a tuple... 

```
receiver(request_finished)((sync_version, async_version))
```

(Or similar)

Tuples are OK until you need to remember which field stands for what. Is it sync, async or async, sync? 

So we have `NamedTuples`, or these days dataclasses: 

```
from dataclasses import dataclass
from typing import Callable

@dataclass
class MyReceiver:
    sync_handler: Callable
    async_handler: Callable

receiver(request_finished)(MyReceiver(sync_handler=sync_version, async_handler=async_version))
```

(Or similar) 

Here the data class is _just_ a namespace for the pair of callables we want to pass. 

If we used a runtime checkable protocol in `receiver`, folks would be free to define a regular class using bound methods, rather than free functions, if the additional class behaviour (state on `self` essentially) was of use. 

(Or something along those lines)

carltongibson:
I think the basic idea here is correct @Arfey here: No-reason at all why built-in signals, and those from third-party apps, shouldn't have a way to opt-in to providing both sync and async versions of the same signal handler. 👍

draft/xxxx-hybrid-signal.md
Arfey: We could leave the receiver decorator untouched and make changes only to the signal API(decorator is only useful for wrapping a function or a class).

```python
request_finished.hybrid_connect(async_handler=async_version, sync_handler=sync_version)
```

However, I prefer using a class-based approach. In my experience, it's much easier to keep things organized this way, sync and async versions often share common logic, and with a class, it's simple to move that into a method with proper encapsulation.

[FieldExtension](https://github.com/strawberry-graphql/strawberry/blob/5bfdf23e9b6c33d5d6d00fbea41214f51d6537d2/strawberry/extensions/field_extension.py#L19C1-L35C10) from the strawberry is a perfect example of how it can look.

carltongibson: Yes, there's no reason to not **allow** people to use a class, especially if there's shared logic as you say.

My proposal was to define a protocol that would also allow named pairs of receivers (which is conceptually what's needed). 

Then you pass either a single callable (as now) or an instance of the protocol (to provide both versions). We already check what kind of receiver was passed at connection, so it's just another branch to handle the both case suggested here. 

Does that make sense? Maybe I'm missing something?

Arfey: That makes sense. I just wanted to emphasize that `receiver` is a decorator that simply calls the connect method under the hood. So, I think we can keep it as it is and just add another method for connecting to the signal.

```python
request_finished.hybrid_connect(async_handler=async_version, sync_handler=sync_version)
```

or

```python
request_finished.hybrid_connect(handler_obj)
```

- we don't need provide dataclass
- we don't need to change default `receiver`'s behaviour

carltongibson: Yes. And kwargs might serve the same purpose of keeping which callable is which clear in the code. 

(These seem like colour of the bike shed questions. We don't have to settle them exactly at this point 🙂)

draft/xxxx-hybrid-signal.md
```diff
@@ -0,0 +1,124 @@
...
+
+When this signal is triggered using `asend`, Django wraps sync functions using
+`sync_to_async`, spawning a thread for each invocation. This adds unnecessary overhead.
+A hybrid signal would allow developers and third-party libraries to register both sync
+and async versions of a handler explicitly. This way, Django can directly dispatch to
+the appropriate function without any cross-context wrapping or thread spawning.
+
+# Specification
+
+A hybrid signal handler can define both sync and async implementations under a single
```
RealOrangeOne, line 56: > A hybrid signal handler **can** define both sync and async implementations

Question: What happens if someone doesn't define both? For example, if a `HybridSignalHandler` only implements `sync_call`, but is triggered in an async context, does Django automatically wrap it in `sync_to_async`? Or should it be validated at definition time that you must define both methods?

Arfey: > What happens if someone doesn't define both?

raise an exception

`HybridSignalHandler` is attempting to remove `sync_to_async` from the request/response cycle.

oliverhaas:
Hello! 

I'm coming here from https://forum.djangoproject.com/t/signals-async-sync-mixed-receivers/45187/3, which is a post of mine where I had the same basic issue this DEP trying to address; I wanted to add no-downside support to packages with signals both for sync and async.

I imagined the API and thus the code in such a package like the (but I really don't mind any of the mentioned suggestions):

```python
@receiver(post_save, sender=MyModel)
def do_something_on_save(sender, instance, **kwargs):
    instance.fun()

if django.VERSION >= (6.2):
    @receiver(post_save, sender=MyModel, pair=do_something_on_save)
    async def ado_something_on_save(sender, instance, **kwargs):
        await instance.afun()
```

Is there any way we could move forward here? I'd be happy to help in any way I can.

codingjoe:
I just reviewed the `Signal.send` and `Signal.asend` implementation. Why do we even need `sync_to_async` here? I think the issue could be resolved without the need for dual implementations.

Maybe there is more to the issue than I am seeing, but it doesn't seem reflected in the problem description just yet.

Arfey:
@codingjoe 

> I just reviewed the `Signal.send` and `Signal.asend` implementation. Why do we even need `sync_to_async` here? I think the issue could be resolved without the need for dual implementations.

Sure, let me show an example. An `ASGI` application uses the sync ORM during a request. At the end of the request, `ASGIHandler` calls `signals.request_finished.asend` to run cleanup. To close the database connection, asend has to be able to run a sync signal receiver. That's where `sync_to_async` comes in.

Does that make sense?

codingjoe:
> Sure, let me show an example. An `ASGI` application uses the sync ORM during a request. At the end of the request, `ASGIHandler` calls `signals.request_finished.asend` to run cleanup. To close the database connection, asend has to be able to run a sync signal receiver. That's where `sync_to_async` comes in.
> 
> Does that make sense?

Oh, so the ORM is the culprit (again). But in that case, you need to run your code in a separate thread anyway. So the example you are giving would benefit from your DEP.

But you are saying all synchronous code that doesn't access the ORM would benefit (if it's implemented correctly), right?

Wild idea: why don't we change the ORM to run on a thread pool by default instead of adding layer after layer of complexity without fighting the root cause?

codingjoe:
@Arfey ah, I see you wrote an async PG backend. I'd believe you'd benefit from a solution from the ORM, or rather the database backends, more. Since Django and its 3rd party packages call `sync_to_async` at every corner, you'll write DEPs 'til kingdom come 🤣

Arfey:
You didn't quite get the problem. Closing the DB connection was just one example because it's in every default Django setup. Any sync signal handler can contain blocking operations: Redis, Celery, DB, Elasticsearch, HTTP calls, etc. That's why every sync handler gets wrapped in sync_to_async. As a result, every Django request spawns a new thread by default. Avoiding that is what this DEP is trying to fix . at least for Django core.

codingjoe:
@Arfey I believe I got that. My point was to make Django not only thread-safe but also async-safe.
Even in Django, all async methods like `aget` only call `sync_to_async`. I think this change will just defer the problem, at least for all first-party and third-party receivers.
And in your own code, you can already write async receivers, which would run on a loop.

I am not per se against the change. I avoid using ASGI mainly because of the ~10ms threading penalty, and your proposal is consistent. Yet I fail to see where this will drive meaningful performance gains. Just like the previous changes with dual implementations. Especially since inside Django most signals are dispatched synchronously.

Please don't take my questioning as disapproval but as passion for the topic. I want to solve this just as much. :)

Arfey:
> I am not per se against the change.

Thank you.

> I avoid using ASGI mainly because of the ~10ms threading penalty

> I appreciate the effort that was put into making this work, but 40ms was in some cases doubling our response times.

ASGI doesn't add 10-40ms latency because of threading itself . it's the sync middlewares. Once you rewrite your middlewares to be fully async, the degradation mostly goes away.

PR #107, opened 2026-04-16
DEP 0019 - Technical governance for the Django project

tim-schilling:
From the motivation section:

This is a revisitation of Django's technical governance in which a simplification and reduction was made to make it more approachable to more people. The goals of these changes are the following:
- Make it easier to enact our governance.
- Make it easier for others to understand our governance.
- Make the governance more flexible, allowing more action with less procedure.

---

Related governance changes:
- https://github.com/django/dsf-working-groups/pull/56
- https://github.com/django/dsf-working-groups/pull/71
- https://github.com/django/dsf-working-groups/pull/72
- https://github.com/django/dsf-working-groups/pull/79


accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+A Minor Change that fails to reach consensus can be escalated to the
+Steering Council for a final decision to merge.
+
+
+How Django's technical direction is determined
+----------------------------------------------
+
+Everyone is encouraged to propose and provide feedback on new features
+for Django at any time on the `new-features GitHub repository`_.
```
benjaoming, line 81: I was wondering how this impacts discussions in the issue tracker and in the forums?

Previously, there has been an informal process where new features proposed on the issue tracker are often deferred for further discussion/feedback in the Forums.

LilyFirefly: We now have a formal process that new features proposed on Trac are redirected to [new-features](https://github.com/django/new-features) instead of the forum. It's been a big success - previously good ideas were getting lost on the forum. With GitHub issues, now have much better visibility of what the open discussions are.

final/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+This document has been placed in the public domain per the Creative
+Commons CC0 1.0 Universal license
+(http://creativecommons.org/publicdomain/zero/1.0/deed).
+
+(All DEPs must include this exact copyright statement.)
+
+.. _DEP 44: https://github.com/django/deps/blob/main/accepted/0044-clarify-release-process.rst
+.. _Mergers Team: https://github.com/django/dsf-working-groups/blob/main/active/mergers-team.md
+.. _Releasers Team: https://github.com/django/dsf-working-groups/blob/main/active/releasers-team.md
+.. _Security Team: https://github.com/django/dsf-working-groups/blob/main/active/security-team.md
```
Stormheg: @tim-schilling is this right? The links 404

Stormheg (commented):
@tim-schilling this file is named `final/0018-technical-governance-6x.rst`, looks like an oversight or typo. This PR title and the DEP contents refer to it as DEP 19.

Stormheg (approved):
I don't know a lot about Django's governance model and thus can't speak to any of the specific details in this DEP, but for what it is worth, I liked reading about how the steering council will be run.

I scanned the referenced DEP 10, and I find this new version much more accessible to read and understand. Thanks for proposing it.

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+
+Steering Council eligibility
+++++++++++++++++++++++++++++
+
+To be eligible to be on the steering council, a person must do the following:
+
+* Be a DSF Individual member.
+
+* Shared any of their corporate affiliations.
```
evildmp: "Have shared"

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+This document has been placed in the public domain per the Creative
+Commons CC0 1.0 Universal license
+(http://creativecommons.org/publicdomain/zero/1.0/deed).
+
+(All DEPs must include this exact copyright statement.)
+
+.. _DEP 44: https://github.com/django/deps/blob/main/accepted/0044-clarify-release-process.rst
+.. _Mergers Team: https://github.com/django/dsf-working-groups/blob/main/active/mergers-team.md
+.. _Releasers Team: https://github.com/django/dsf-working-groups/blob/main/active/releasers-team.md
```
fadamakis:
```suggestion
.. _Releasers Team: https://github.com/django/dsf-working-groups/blob/main/active/releasers.md
```

final/0019-technical-governance-6x.rst
RealOrangeOne: It will be - the Security Team WG doc is still being worked on.

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+Everyone is encouraged to propose and provide feedback on new features
+for Django at any time on the `new-features GitHub repository`_.
+
+For features which qualify as a Major Change, proposers may be asked to
+use `the DEP process`_. Likewise, other changes, for example to processes,
+that do not fit the new features description may be presented as DEPs
+
+If discussion of a Minor Change has failed to produce consensus, a
+member may ask the Steering Council to make a decision.
```
shaib: "a member" of what? or should this be "a paticipant in the discussion" or "anyone"?

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+Rejected discussions and features are eligible to be revisited at the
```
shaib:
```suggestion
Rejected suggestions and features are eligible to be revisited at the
```

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+* Have three or more Steering Council Qualities.
+
+* Not be a members of the `DSF Board`_.
+
+Steering Council Qualities
+**************************
+
+Below are several traits a Django contributor may possess that are beneficial
+when on the Steering Council. An individual is unlikely to possess all of
+them, but qualified candidates should have more than one. Each trait is
```
shaib: "more than one" or "three or more" (above)?

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+We want people who know how Django is used to build real-world applications.
+The Steering Council needs to understand the needs of web applications and
+experience with actual usage allows that.
+
+Indicators:
+
+* Experience building and maintaining production Django apps over years
+
+* Has a network of people who maintain and build Django applications
```
shaib, line 276: I find this unclear, or maybe just too abstract.

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+We want people who have a strong understanding of Django from the framework's
+perspective. The Steering Council needs to be able to weigh technical
+decisions appropriately and familiarity with the code supports that.
+
+Indicators:
+
+* Familiar with several areas of the Django codebase
+
+* Has helped maintain Django or another well-used project for years
```
shaib:
```suggestion
* Has helped maintain Django or a well-used project in the Django ecosystem for years
```

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+* It is determined that they did not possess the qualifications of a
+  member of the Steering Council. This determination must be made
+  jointly by the other members of the Steering Council, and the DSF
+  Board. A valid determination of ineligibility requires that all
+  other members of the Steering Council vote to declare the affected
+  person ineligible and that all members of the DSF Board who can
+  vote on the issue (the affected person, if a DSF Board member, can
+  not vote) vote "yes" on a motion that the person in question is
+  ineligible.
```
shaib:
```suggestion
  person ineligible and that all members of the DSF Board 
  vote "yes" on a motion that the person in question is
  ineligible.
```
For the qualification to be needed, a person should be able to serve on both DSF Board and Steering Council simultaneously, and this is not the case.

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+statement of no-confidence in the Steering Council by identifying a
+material breach of their duties as defined in the technical governance.
+Upon seconding by another individual member of the DSF the DSF Board
+will evaluate the merits of the statement of no-confidence in their
+next meeting.
+
+If the statement is found to be accurate and correct the DSF Board shall
+inform the Steering Council of the breach and provide 2 weeks to
+rectify said breach. If the Steering Council fails to rectify the
+breach in the time allotted, a new Steering Council election will be
```
shaib: It should probably be specified that the Board arbitrates if the breach was rectified.

final/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+  process.
+
+In the event that the Steering Council feels the Security Team has
+used the above powers inappropriately, the Steering Council may appeal
+to the DSF Board to mediate the issue. Any member of the DSF Board who
+is also a member of the Security Team will
+abstain from participation in the DSF Board's decision-making in such
+mediation. The decision of the DSF Board in the dispute will be
+binding on both the Steering Council and the Security Team.
+
```
shaib: Is this special to the Steering Council? Doesn't it hold for any team, institution, or even individual?

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+
+Decision making process of the Steering Council
++++++++++++++++++++++++++++++++++++++++++++++++
+
+When asked to make a technical decision, the Steering Council should
+first discuss this amongst themselves. If there's agreement on a course
+of action, a single member will respond on the `Django Forum`_, ticket, or
+`new-features GitHub repository`_ on behalf of the Steering Council. It
+may optionally include a dissenting opinion if someone wishes to include one.
```
shaib, line 149: This paragraph defines how to handle and publish decisions of the Steering Council, but not how to make them. It essentially assumes rough consensus (within the council) on any issue.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+  governance documents (DEP 10 & 12).
+
+  * A single document avoids having to read two documents to
+    understand the governance.
+
+* Reducing the number of sections and topics.
+
+  * DEP 10 sought to provide information on how the new governance
+    would operate. While helpful in that moment, it eventually
+    became a burden when revisiting the document.
```
shaib, line 623: These together suggest that we should differentiate between documents which are literal DEPs -- documents describing the suggested change -- and documents describing the "achieved state", a "constitution" of sorts.

DEP 12, in particular, was mostly a change document. This one appears to be more of a result document.

final/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+  * The community and `DSF Board`_ both have mechanisms to remove an unruly
+    Steering Council. These limitations appear to have only limited
+    productive functions of the Steering Council. Allowing approving
+    DEPs during elections allows for the community to operate more
+    freely. It's possible that a newly elected Steering Council will
+    change decisions from the previous Steering Council. This is seen
+    as the system working. If at any point, the trust in this system
+    is abused, the community can rely on reporting members for
+    violations of Django's Code of Conduct.
```
shaib: Are all possible abuses code-of-conduct violations?

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+    Conduct Violation.
+
+  * This language was changed from vetoed to rejected to better match
+    the decision making process for the Steering Council.
+
+* Not allowing Board members to serve on the Steering Council.
+
+  * This simplifies some of the governance by reducing exception cases.
+    Plus, these roles are meant to be time-intensive and it's unlikely
+    for someone to do both well.
```
shaib: What were the exceptions?

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+    for someone to do both well.
+
+* Removing the explicit voting from the decision making of the Steering
+  Council.
+
+  * This accurately reflects the process the 6.X Steering Council has
+    been using. Future Steering Councils may internally decide to use
+    voting, however it is not a necessary process. The discussion and
+    deliberation is a more accurate representation of how the Django
+    community arrives at consensus.
```
shaib, line 700: But this is not about how the Django Community makes decisions, it's about the Steering Council specifically; and at least in some cases, the Council's role is to decide when a decision needs to be made and consensus cannot be achieved.

I think that whenever a decision is made without consensus, it is preferable to have transparency (of course, consensus is inherently transparent).

shaib (commented):
Generally, I find this document not clear enough about changes it makes. As an example, "reducing the scope of eligible voters" -- the doc claims it "allows effectively the same number of people to participate", but it never mentions if it's just the same number or the same people, nor what were the previous qualifications.

As such, it may be a good spec, but I find it lacking as an enhancement proposal.

Further point-by-point comments inside.

final/0019-technical-governance-6x.rst
Stormheg: @RealOrangeOne I was referring to all teams mentioned here. I can see the PR to add the security team documentation is still open.

There are already mergers and releaser team documents in the repo, but they are available under a different filename. Like https://github.com/django/dsf-working-groups/blob/main/active/releasers.md

accepted/0019-technical-governance-6x.rst
benjaoming: Got it! I was unsure how the new proposal affects the existing patterns. Makes sense to say that it just replaces anything that previously existed 👍 

Will see how the discussions unfold on new-features. One concern could be if small tracks for detailed decisions in each issue on the new features need some more structure to work. But maybe that can be solved iteratively through the guidance provided in the issue templates.

benjaoming: GitHub is failing to mark this thread as resolved, but please consider it as resolved.

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+To be eligible to be on the steering council, a person must do the following:
+
+* Be a DSF Individual member.
+
+* Shared any of their corporate affiliations.
+
+* Have three or more Steering Council Qualities.
+
+* Not be a members of the `DSF Board`_.
```
benjaoming:
```suggestion
* Not be a member of the `DSF Board`_.
```

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+teams, please see the `django/dsf-working-groups GitHub repository`_.
+
+
+Interaction of the Steering Council and the Security Team
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+The `Security Team`_ has the following powers:
+
+* To request a Merger merge code to fix a security issue being
+  handled under Django's security process.
```
benjaoming:
```suggestion
* To request a Merger merges code to fix a security issue being
  handled under Django's security process.
```

I'm a bit unsure why _requesting_ a merge is considered a "power"? Anyone can request something? The power would be to approve, merge or demand?

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+* To request a Releaser issue a release of Django containing code to
+  fix a security issue being handled under Django's security
+  process.
```
benjaoming:
```suggestion
* To request a Releaser issues a release of Django containing code to
  fix a security issue being handled under Django's security
  process.
```

Same question for this point, what does "request" really entail here?

benjaoming:
Can/should this mention something about the governance of the `new-features` repo itself? The repo has a whole process that someone should own/develop? Is that the Steering Council?

accepted/0019-technical-governance-6x.rst
carltongibson: > but not how to make them

Yes, this is deliberate. The specification of the voting mechanism is DEP 10 was as good a demonstration as we could hope for of why specifying such a thing (especially for a small organisation like ourselves) is too heavy an approach. 

For me (personally) _rough consensus_ is exactly what we should be aiming for. A loose show of hands is one thing, but (for me) a formal vote marks a failure. (That we resorted to something so crude, even between just the five of us, is a lack of… something… I would argue.) 

**But** equally, other SCs in the future can make their own decisions. Simple majority, super majority, unanimous, are all reasonable options to follow: it very much should be up to each SC to decide how they want to proceed.

The point with a _dissenting opinon_ is exactly to give voice in any circumstance where (effectively) _unanimous_ didn't carry the day. 

In reality, despite often starting a discussion from quite different positions, we've not had a case where this has been necessary. Legislating for it (too heavily) is precisely what we're aiming to avoid with this new document. 

(We can clarify here, if needed, sure)

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+Copyright
+=========
+
+This document has been placed in the public domain per the Creative
+Commons CC0 1.0 Universal license
+(http://creativecommons.org/publicdomain/zero/1.0/deed).
+
+(All DEPs must include this exact copyright statement.)
+
+.. _DEP 44: https://github.com/django/deps/blob/main/accepted/0044-clarify-release-process.rst
```
codingjoe: This link didn't work for me.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+* To create/update `teams`_
+
+Steering Council's goals
+++++++++++++++++++++++++
+
+The Council's goal is twofold - to safeguard big decisions that affect
+Django projects at a fundamental level, and to help shepherd the
+project's future direction.
+
+While the Council should not define this direction entirely by itself,
```
codingjoe, line 133: The “should not” phrase is relatively vague. My understanding is that most decisions are not self- but community-initiated.

IMHO, it may be a fairly good practice to separate those two powers.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+The `Security Team`_ may request a Security Release of Django. This
+should be performed at or as close as is practicable to the time of
+release requested by the `Security Team`_.
+
+
+Steering Council role
+---------------------
+
+The Steering Council provides oversight of Django's development and
```
codingjoe, line 109: A few lines above the release processes defined by DEP0044. If this isn't a conflict, which I presume, it might need clarification.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+release process, assists in setting the direction of feature
+development and releases, selects Mergers and Releasers, and has a
+tie-breaking vote when other decision-making processes fail.
```
codingjoe, line 112: Hm… is that power limited to technical decisions? It seems odd that the steering committee would have the deciding vote on governance. I'd presume this power would be with the elected board.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+The Steering Council consists of five members.
+
+The powers of the Steering Council are:
+
+* To make a binding decision regarding any question of a technical
+  change to Django.
+
+* To manage the Steering Council's membership via an election with the
```
codingjoe, line 121: In practice it might not make a big difference, but it's always good if a body with this much power doesn't self-elect members. If the governing body is appointed by the board, which is elected, you end up with a stronger legitimacy.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+  `DSF Board`_.
+
+* To create/update `teams`_
```
codingjoe, line 124: As I understand it, it also delegates powers and responsibilities to specialized teams, right?

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+
+Steering Council's goals
+++++++++++++++++++++++++
+
+The Council's goal is twofold - to safeguard big decisions that affect
```
codingjoe, line 129: emdash 🤖 
```suggestion
The Council's goal is twofold — to safeguard big decisions that affect
```

final/0018-technical-governance-6x.rst
codingjoe: Or "disclosed"

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+Steering Council eligibility
+++++++++++++++++++++++++++++
+
+To be eligible to be on the steering council, a person must do the following:
+
+* Be a DSF Individual member.
+
+* Shared any of their corporate affiliations.
+
+* Have three or more Steering Council Qualities.
```
codingjoe, line 161: That seems highly subjective. As much as I believe agreeing on a common set of values is helpful, they make for weak selection criteria, which in turn takes legitimacy from the selection process and opens it to subjective criticism.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+help communicate progress.
+
+* The Steering Council's meeting minutes can be found in the
+  `steering-council GitHub repository`_.
+
+* The `Django features roadmap`_ for can be found in the
+  `new-features GitHub repository`_
+
+
+Teams
```
codingjoe, line 459: They are everywhere else referred to as working groups. Sticking to one term might simplify comprehension of DSF governance.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+   Forum and any other appropriate venues. The five candidates with the
+   highest vote totals will immediately become the new Steering Council.
+
+Removing a single Steering Council member
+*****************************************
+
+A member of the Steering Council can be removed in the following
+ways:
+
+* They become ineligible due to actions of the Code of Conduct
```
codingjoe, line 387: The CoC currently has a working group, which otherwise isn't mentioned here. Which might be important if they can dismiss a member.

codingjoe (commented):
Thanks for the hard work, Tim! I agree with most points, but for the sake of improvement, my comments will focus on where I disagree:

I was pondering for a while if it was my place to say something, but I am not in love with the “traits.” I think I grasp their purpose, but I think their subjectiveness weakens the steering committee and, in turn, Django's technical oversight.

Moreover, I am missing accountability. I love how all meetings and decisions are currently published. Let's make this wonderful practice a written requirement in the future.

I'd even go a step further and propose a transparency amendment, allowing every DSF member to make a freedom of information inquiry, which the committee must always respond to in a timely manner. I don't believe there is currently much use for it, but it's a remarkable tool to build trust.

Lastly, I don't believe the steering committee should be self-governing. At least not on paper. I'd prefer a fully board-appointed and dismissed committee. This way the member board vote and its inherent legitimacy would propagate unhindered to the steering committee.

tim-schilling:
> Can/should this mention something about the governance of the new-features repo itself? The repo has a whole process that someone should own/develop? Is that the Steering Council?

@benjaoming I think the mentions of the repository are sufficient at this time. We're still iterating on what that process looks like and sorting things out there. I think this document is better directing people to that repo to understand the new features process as a whole rather than trying to capture it here (and creating a second place for it to be updated in the future).

Yes, it's currently owned by the Steering Council. We've had discussions on how to expand that membership, but nothing conclusive. It's going to be a focus for the second half of the year.

accepted/0019-technical-governance-6x.rst
tim-schilling: FYI, this language is taken directly from [DEP 12](github.com/django/deps/blob/main/final/0012-steering-council.rst#:~:text=twofold).

final/0018-technical-governance-6x.rst
tim-schilling:
```suggestion
them, but qualified candidates should have three or more. Each trait is
```

final/0019-technical-governance-6x.rst
tim-schilling: Yup yup, but this section is discussing the relationship between these two groups specifically, that's why it mentions them explicitly.

final/0018-technical-governance-6x.rst
tim-schilling: Anywhere in DEP 10 where it had to mention something about a person being both a board member and steering council member. For example, the exception case that we missed, but you caught elsewhere in the document: https://github.com/django/deps/pull/107#discussion_r3104920668

final/0018-technical-governance-6x.rst
tim-schilling:
```suggestion
.. _DEP 44: https://github.com/django/deps/blob/main/final/0044-clarify-release-process.rst
```

final/0018-technical-governance-6x.rst
tim-schilling: I hear you. I can see why it can feel off.

The security team doesn't have the actual permission to merge or release code. They need the mergers or releasers to do that. Using the language of request rather than demand is a bit more open to allowing leeway in how things go.

If we're running into a situation where teams are nitpicking the language here saying, "you can only request!!!" we're running into major problems. I think it's fair to assume people in our community are going to be collaborative. If people are wielding specific language to shield their behavior repeatedly, that's a CoC concern.

accepted/0019-technical-governance-6x.rst
tim-schilling: I recognize that's an oddity, but it seems small and reasonable to deal with for the time being.

accepted/0019-technical-governance-6x.rst
tim-schilling: committee == working group in this sense. We can revise the language to be consistent. Suggestion here: https://github.com/django/deps/pull/107/files#r3110813593

final/0018-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+   highest vote totals will immediately become the new Steering Council.
+
+Removing a single Steering Council member
+*****************************************
+
+A member of the Steering Council can be removed in the following
+ways:
+
+* They become ineligible due to actions of the Code of Conduct
+  committee of the DSF. If this occurs, the affected person
```
tim-schilling:
```suggestion
  Working Group of the DSF. If this occurs, the affected person
```

accepted/0019-technical-governance-6x.rst
tim-schilling: @codingjoe I feel like you may have a misunderstanding on how the process works based on a few of your comments. I'm going to explain a bit more in this comment, but I apologize if I'm over explaining. The Steering Council is elected by the DSF individual members. The candidates who are able to stand for election are limited by these traits which will be evaluated by the DSF board. People are able to vote for whoever they want.

Over the last few years, we've made consistent efforts to broaden the candidates for the Steering Council. By expressing traits, we're hoping people identify with several of them and realize that they are qualified to self-nominate themselves. The goal here is to be more accessible to candidates who haven't seem themselves as eligible in the past.

accepted/0019-technical-governance-6x.rst
tim-schilling: To be clear, manage is referring to removing members or adding up to 2 replacement members if others step down. The Steering Council as a whole is elected by the DSF individual members, facilitated by the DSF board.

accepted/0019-technical-governance-6x.rst
tim-schilling: Please review my other comments. I think this one may be resolved since the SC is also elected.

accepted/0019-technical-governance-6x.rst
tim-schilling: Can you elaborate on the confusion please?

final/0018-technical-governance-6x.rst
tim-schilling: Yes, that's how I interpret this. Any community member can call for help unsticking a lack of consensus.

accepted/0019-technical-governance-6x.rst
codingjoe: @tim-schilling no worries, I wrote my comments as a complete newcomer to the DSF governance. So you can't overexplain anything :) My aim was to provide a fresh set of eyes and I appreciate your long explanation.

Back to the topic: I see, that's a great effort. But seeing how I was looking at this from the outside in and confused it, maybe this can be phrased under the pretense of inclusion. This particular sentence makes it seem more exclusive to me. Not a native speaker though.

final/0018-technical-governance-6x.rst
benjaoming: I 100% agree with the sentiment about having a collaborative atmosphere without need for strong legal phrases :+1: 

I'll try to put a suggestion in the introduction to the list. That way, these two can remain small grammar nit-picks :blush:

final/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,775 @@
...
+changes to the technical governance document.
+
+To see the list of teams and the process of adding new
+teams, please see the `django/dsf-working-groups GitHub repository`_.
+
+
+Interaction of the Steering Council and the Security Team
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+The `Security Team`_ has the following powers:
```
benjaoming:
```suggestion
The `Security Team`_ ensures that security-related changes are addressed with the appropriate imminence and priority by Mergers and Releasers. The Team's role is:
```

okay I know it's a nit, but it's to avoid the word "power" and add a bit more verbosity.

See @tim-schilling's [comment below](https://github.com/django/deps/pull/107/changes#r3110773054).

accepted/0019-technical-governance-6x.rst
codingjoe: This was regarding the release process. I was under the impression that this is already formalized in DEP 44. However, further below, there are more examples of what this mains. I just didn't understand it at this point reading. I think this comment is resolved.

accepted/0019-technical-governance-6x.rst
codingjoe: Thanks. I hope I am not making too much trouble. Please disregard my comments if they don't advance the purpose of this change.

jacobtylerwalls:
> I'd prefer a fully board-appointed and dismissed committee. This way the member board vote and its inherent legitimacy would propagate unhindered to the steering committee.

Responding briefly to this (and not sure if this idea has already been withdrawn) -- I appreciate the separation of concerns by having the DSF Board not in charge of technical matters but rather concentrated on the practicalities of running a non-profit. Obviously there is overlap (it's a nonprofit with a technological mission), but my understanding is that it's already a problem that people run for the DSF Board thinking that a history of code contributions is an important qualification for the role, when really they should be running for the steering council instead.

codingjoe:
> > I'd prefer a fully board-appointed and dismissed committee. This way the member board vote and its inherent legitimacy would propagate unhindered to the steering committee.
> 
> Responding briefly to this (and not sure if this idea has already been withdrawn) -- I appreciate the separation of concerns by having the DSF Board not in charge of technical matters but rather concentrated on the practicalities of running a non-profit. Obviously there is overlap (it's a nonprofit with a technological mission), but my understanding is that it's already a problem that people run for the DSF Board thinking that a history of code contributions is an important qualification for the role, when really they should be running for the steering council instead.

Thanks, @jacobtylerwalls. Tim has addressed this comment, and it was a misunderstanding on my side. Of course the board shouldn't be responsible for the technical oversight. My understanding is now that, under normal circumstances, the steering committee is member-voted. Making my comment obsolete.

As previously mentioned. I read this as a complete noob to provide a fresh set of eyes. I hope it doesn't stir up too much trouble if I'm getting things wrong here and there 🙈

final/0018-technical-governance-6x.rst
shaib:
```suggestion
If discussion of a Minor Change has failed to produce consensus, anyone
may ask the Steering Council to make a decision.
```

accepted/0019-technical-governance-6x.rst
shaib: This is a laudable position -- a model to follow, even -- but I think the wording that hides it behind the missing "else" of "if there's agreement" is suboptimal.

```suggestion
When asked to make a technical decision, the Steering Council should
first discuss this amongst themselves. It is up to each Steering Council
to decide how to resolve disagreements. When there's a decision on a course
of action, a single member will respond on the `Django Forum`_, ticket, or
`new-features GitHub repository`_ on behalf of the Steering Council. It
may optionally include a dissenting opinion if someone wishes to include one.
```

One thing to consider around this is transparency -- the legislated voting system did, to a degree, give us that. When I go voting for the next SC, and some existing members are running again, I want to have some info about how they functioned within the SC. On the other hand, there is value in keeping internal discussions private -- that usually makes it easier for people to both express opinions and change them. I don't have a concrete suggestion how to solve this tension, but I thought it worth bringing it up.

final/0019-technical-governance-6x.rst
shaib: My point is that there is no special relationship between these two groups.
I think the section is redundant -- The "Interaction with other teams" section below covers the Security Team as well.

final/0018-technical-governance-6x.rst
shaib:
```suggestion
  * This simplifies some of the governance by reducing exception cases
    related to serving on both simultaneously.
  * Plus, these roles are meant to be time-intensive and it's unlikely
    for someone to do both well.
```

accepted/0019-technical-governance-6x.rst
shaib: Let's keep this discussion in one place, where the decision process is defined.

shaib (commented):
This is all responses to previous comments.

final/0019-technical-governance-6x.rst
tim-schilling: I think for what we would see in the case of passing DEPs during an election that the community at large doesn't want, yes. That combined with the oversight from the board for governance changes makes this safe enough from our perspective.

accepted/0019-technical-governance-6x.rst
carltongibson: > One thing to consider around this is transparency ...

Sure, and transparency is important, and something we've tried to promote via [publishing our regular meeting notes](https://github.com/django/steering-council), actively engaging on topics that come up in the Forum and new-features repo, and publishing the more occasional [Update](https://www.djangoproject.com/weblog/2026/feb/11/steering-council-2025-year-in-review/)™. 

That goal is hopefully addressed by the `Steering Council reports` section of this document. (To which clarifications are of course welcome.)

A separate thing is whether SC deliberations (rather than decisions) should be public, and/or constrained to a specific format. On both of those, I'd lean to a hard no (for the reasons given). 

How do we judge an SC's performance? Well, what got done during their term? Did they achieve what they set out to? Did they do what they said they would? Or at least try? 

One big change is that ≈ DEP 10 conceived of the SC as ≈ only there for tie-breaks, whereas it became evident that the SC needs to play a more active role than that. In which case, I think (both) that if an individual SC member is sat doing nothing, that would be pretty evident in a community like ours, and that public formal votes wouldn't help that. (Indeed, the latter would likely give an appearance of activity where none really exists.) 

Or, my thoughts are along those kind of lines. It's not the _How decisions are made_ bit that's the key locus there. 🤷‍♂️ 

(No problem with the suggested edit here!)

final/0019-technical-governance-6x.rst
shaib: I find the "oversight of board" part convincing that the change is for the better. I find references to the Code of Conduct in this context counter-productive and even dangerous, because, IMO, they widen the scope of code-of-conduct enormously.

If a Steering Committee approves a DEP during elections because they think it is popular amond DSF members, although they believe it will hurt the project -- "election bribe" -- that is a bad move, but not a CoC violation. Even if a controversial DEP gets approved because it serves the interests of Council members who think they will be replaced, that is IMO still not a CoC issue (and the CoC Committee, last I checked, has authority to suspend people, but not to overrule DEP approvals).

In short -- I'd much rather that lack-of-integrity not be seen, in and of itself, as a CoC issue.

accepted/0019-technical-governance-6x.rst
shaib: | It's not the _How decisions are made_ bit that's the key locus there. 🤷‍♂️

Well, the title of this section is "Decision making process of the Steering Council"... 

| How do we judge an SC's performance? ...

That's not my question, though. I expect Steering Council members to have personal agendas, sometimes-strong opinions about the directions the project should evolve in, and I don't really expect those to be fully expressed in their public declarations (they might express them on the forum and in PR reviews, but it's very easy to miss such posts). Say I like "enterprise" features, and a specific SC member keeps convincing others to approve such features -- will I know to vote for them next time?

If the existing posts had, for every decision taken, a "Initial positions / arguments / final positions" summary, that would probably give all the info I want, but I'm well aware of the load this would put on whoever is doing the current meeting notes (and maybe even on the discussions themselves), so I don't think requiring that is a good idea.

As I said, I don't have concrete suggestions to make this better. There seems to be a trade-off here, and as such, the point selected along it is fine. I just wanted the point discussed in case someone does come up with an improvement.

accepted/0019-technical-governance-6x.rst
tim-schilling: Sounds good, thanks @codingjoe

accepted/0019-technical-governance-6x.rst
tim-schilling: @codingjoe I'm going to mark this as resolved. Please let me know if you think it shouldn't be.

accepted/0019-technical-governance-6x.rst
tim-schilling: @codingjoe I'm going to mark this as resolved. Please let me know if you think it shouldn't be.

accepted/0019-technical-governance-6x.rst
tim-schilling: That's correct. I think part of creating/updating a team should imply potential delegation of responsibilities.

accepted/0019-technical-governance-6x.rst
tim-schilling: I'm going to mark this as resolved. If there's no difference between the suggestion and the existing language, I'd like to reuse the existing language.

accepted/0019-technical-governance-6x.rst
tim-schilling: > The “should not” phrase is relatively vague. My understanding is that most decisions are not self- but community-initiated.

The subsequent sentences in this paragraph outline specific actions the Steering Council can take to help encourage and support community driven ideas. I feel like that should help dispel the vagueness for a reader.

> it may be a fairly good practice to separate those two powers.

Can you clarify which two powers you're referring to here please?

accepted/0019-technical-governance-6x.rst
tim-schilling: > ... I was looking at this from the outside in and confused it, maybe this can be phrased under the pretense of inclusion. This particular sentence makes it seem more exclusive to me.

Do you have a suggestion on how to reword this? I'm not sure how to define a requirement as inclusive. By its definition it is gatekeeping.

accepted/0019-technical-governance-6x.rst
tim-schilling: Let me provide some context and see if we can arrive at a better way to phrase this. If a person has been building python apps for a while, but not always Django applications, they likely have a decent amount of knowledge of Django applications. If that person also has a group of people they can rely on for questions about Django or how to build things, that means their accessible knowledge about Django is more than what they know themselves. They would be able to survey their network to ask for opinions or understand a different concept more readily.

We wanted to capture that distinction because many people will just think about what do they themselves know rather than, who is in my direct circle that I could talk to to understand this? That latter has a lot of benefit at the Steering Council level because the surface area is so large.

Let me know where there's confusion and/or how we can reduce that concept into a short sentence.

Maybe: 

> Has a network of Django practitioners and builders they can consult to access perspectives and expertise beyond their own

final/0018-technical-governance-6x.rst
tim-schilling:
```suggestion
rectify said breach. If the Steering Council fails to rectify the
breach to the Board in the time allotted, a new Steering Council election will be
```

final/0019-technical-governance-6x.rst
tim-schilling: I believe the implication of this information is distinct.  The Security Team is held in a different regard according to our past governance and we're maintaining that for now. There's a difference between the Security Team requesting a release and the Accessibility Team requesting a release.

However, your point that the interaction between the SC and the teams covers how the security team and SC should interact is a good one. The delegation of requesting a release of Django for a security concern would be better defined in that team's charter.

Thoughts?

(@benjaoming, I think this would resolve your other comments as well by removing the text altogether)

accepted/0019-technical-governance-6x.rst
tim-schilling: Yes, exactly. This is a replacement of DEP 10 and 12 so a person wouldn't have to sort out which pieces are being overridden.

I believe you're pointing out, what do we do in the future when we need to change this again and that's a good question. Maybe we pull this into the Django code base so there's a source of truth somewhere and then allow the DEPs to be change documents for governance.

We could also host it on the djangoproject.com/foundation/ section of the website so it doesn't need to impact the codebase.

final/0019-technical-governance-6x.rst
tim-schilling: I understand the concern around wanting to limit the impact of a wanton, ill-intentioned Steering Council. However, let's keep in mind our community is fairly small and they are elected. I think we should strive for governance that helps community members be effective and assume members act with positive intentions until we find otherwise.

There are four safeguards (I missed two from before):
- The election process itself
- The DSF Board takes action on their own
- A CoC violation (effectively our outside arbitration)
- Vote of no-confidence by a community member

accepted/0019-technical-governance-6x.rst
tim-schilling: @shaib While you didn't ask this directly, when considering folks for re-election, you can look at what the team has accomplished, then dig into those to see how those accomplishments came together.

And for what it's worth, this SC has commented on various topics with their views before the team makes a decision. Though that doesn't make it easy to find when the next election cycle comes up.

final/0018-technical-governance-6x.rst
tim-schilling:
```suggestion
* Have shared any of their corporate affiliations.
```

accepted/0019-technical-governance-6x.rst
tim-schilling: @shaib I'm circling back on things. Can you let me know if the above clarification helps please?

final/0019-technical-governance-6x.rst
tim-schilling: @benjaoming I'm circling back on things. Can you let me know if the above is agreeable to you?

accepted/0019-technical-governance-6x.rst
tim-schilling: @codingjoe I'm circling back on things. Are there specific changes you're wanting to see here yet?

accepted/0019-technical-governance-6x.rst
tim-schilling: I believe this has been resolved in the wording changes.

tim-schilling:
When folks have time, can you review the open threads and see if there are any loops that can be closed?

accepted/0019-technical-governance-6x.rst
shaib: @tim-schilling sorry -- I had marked this as still requiring attention, but then lost track of it.

The clarification doesn't help, and it's my fault, for being, eh, too abstract...

The problem I see is that the list of traits and indicators is given as a way to judge eligibility of candidates, but that has two different perspectives:
- As a potential candidate, am I eligible?
- As a voter looking at a candidate, are they eligible?

All the other indicators (I've just gone through the list to verify again) are based on work which usually leaves visible evidence, and so, are easy for a candidate to claim and a voter to verify. This one, not so much. 

The explanation in your message clarifies that, indeed, when drafting this indicator, you had the "am I eligible" angle in mind; but the other one is missing.

I'm pointing out, just for completeness, that the list of Qualities, in its own presentation, is listed as "beneficial", and "if you feel you possess that trait [...] that is sufficient" -- i.e. mostly for self-assessment; but having three qualities is listed before as an eligibility criterion on the same level as the objective "Be a DSF Individual member" and "Not a DSF Board member".

So, basically, I feel a tension between the apparent subjectivity of this indicator, and the objective role that I thought indicators should fulfill (in addition to a subjective role).

final/0019-technical-governance-6x.rst
shaib: I failed to make myself clear:

The text we're discussing here is part of an explanation of  a change this DEP introduces (elsewhere), allowing the approval of DEPs during SC elections. I think the change is good, and I have no issues with it.

In explaining this change, a reference is made to the CoC committee, which, IMO, widens the scope of what's considered a CoC violation. This part is supposed to be just explanatory, but is still language in a DEP, and as such, will be accepted as authoritative. I would prefer if the reference to the CoC be removed from this explanation.

accepted/0019-technical-governance-6x.rst
tim-schilling: I think I understand. You're raising the concern that a person could claim this trait, but the general public wouldn't be able to tell. If I'm understanding you properly, I do agree there's a tension there. Though I think we need to tolerate it until we have a good reason not to.

I suspect you're coming at this from the angle of "how do I as a voter know if a person meets the criteria?", I'd point out, it's the candidate's responsibility to explain that in their nomination statement (and to the DSF board when nominating themselves). If they don't communicate their utility to the community, there's not much that you as the a voter can do to help them.

tim-schilling: Hi @shaib, the SC discussed this and landed on keeping this as is. I appreciate you highlighting the difference in subjectivity for this indicator. Given that the indicators are an incomplete set of examples of a way a person could show they have a quality, we're inclined to keep it. I'm going to mark the conversation as resolved, but you're welcome to follow-up with additional questions.

final/0019-technical-governance-6x.rst
tim-schilling: Hi @shaib, we discussed this in the meeting and considered removing it, but decided to keep it as is. We felt that the CoC WG already had this ability and we're just highlighting it here. It also factored in that this is in the rationale section and not the "constitution" piece. Related to that, it's important for us to include why we are comfortable removing the limitation of processing DEPs during an election. Without that reference, this change would appear that we're removing that limitation without concern for possible inappropriate SC actions. I'm going to resolve this conversation, but please feel free to follow-up with additional questions!

final/0019-technical-governance-6x.rst
tim-schilling: Thanks @Stormheg, I fixed the links.

final/0019-technical-governance-6x.rst
tim-schilling: The section about interaction with the security team was removed entirely. Thank you @shaib for making the doc even shorter! 🥳

final/0019-technical-governance-6x.rst
tim-schilling: This section was removed.

accepted/0019-technical-governance-6x.rst
tim-schilling: @codingjoe I'm going to resolve this conversation.

accepted/0019-technical-governance-6x.rst
tim-schilling: Marking this as resolved to match the decision to leave this as is.

accepted/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,761 @@
...
+  writing, of the nomination. The DSF Board will check the
+  qualifications of the person nominated, and the DSF Board
+  will notify the Steering Council of the result. If the DSF
+  Board determines the nominated person is not qualified, the
+  nomination must be discarded.
+
+* A qualified nominee will fill the vacancy if the Steering Council
+  votes to appoint the nominee. As an exception to the Steering
+  Council voting process described above, this vote must be completed
+  within one-week and it must be unanimous approval.
```
LilyFirefly, line 425: What happens in the case that one member is temporarily unavailable? For example they might be in hospital for that week.

Would we defer starting the vote until they are available?

LilyFirefly (approved):
Looks good to me! I have one question about a (hopefully!) rare edge case, but it's not a blocker for me.

accepted/0019-technical-governance-6x.rst
tim-schilling: If we were facing that situation now (4 members trying to appoint a new person and one of the four is indisposed), that's what I would suggest. Unless there was something more urgent that required us to have the new person on board. If that was the case and this was getting in the way of us being productive of urgent matters, I'd likely push for tweaking the governance to not get in the way.

accepted/0019-technical-governance-6x.rst
shaib: Hi @tim-schilling , thanks for your (and the SC's) understanding and consideration.

final/0019-technical-governance-6x.rst
shaib: Apparently (ref also [the discussion in the security team charter PR](https://github.com/django/dsf-working-groups/pull/56#discussion_r3419065901)) my interpretation of the scope of the CoC is different from that of the SC. 

@django/coc-committee do you see it as a CoC violation, if a person in a technical role knowingly promotes a technical action which isn't in the best interest of the project -- doing so politely, communicating respectfully and not making anyone feel personally uncomfortable?

carltongibson: I'm going to unresolve this (at least for now): It seems, both here and on the security team thread, that there's something to clarify. 

The **intent** is that CoC applies to everyone: Board, Steering Council, and Security Team included. (I assume we're all agreed on that. And it's what I take to be the _status quo ante_.) So, CoC violations can lead to suspension or removal from any of those positions. Maybe the CoC team needs/wants support from other groups in such a circumstance. No problem there. (I assume we still all agree?)

If a community member had a concern about a polite but technically egregious SC lame-duck decision, either we head into "The Next SC reverses it" territory OR they ask the Board to review.

The CoC case applies (one imagines) if the lame-duck SC (somehow) goes off the rails. (Whatever that means: no doubt obvious when you see it.) — So there are three mechanisms in play: Next SC, Board Review, and CoC. The latter only applies in one kind of case. 

The (only) point (I think) of this mention here is just to (re)clarify that the usual guardrails continue to apply, and that DEP approval to continue during the pre-election phase is not problematic, at least in part, for that reason.

We were/are keen to maintain that thought. 

🤔

final/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,761 @@
...
+
+  * The community and `DSF Board`_ both have mechanisms to remove an unruly
+    Steering Council. These limitations appear to have only limited
+    productive functions of the Steering Council. Allowing approving
+    DEPs during elections allows for the community to operate more
+    freely. It's possible that a newly elected Steering Council will
+    change decisions from the previous Steering Council. This is seen
+    as the system working. If at any point, the trust in this system
+    is abused, the community can rely on reporting members for
+    violations of Django's Code of Conduct.
```
carltongibson: @tim-schilling @shaib Would a rephrase along these lines maintain what the SC was keen to express, whilst alleviating concerns?: 

```suggestion
* Allowing approving DEPs during elections.

  * This is about not putting barriers in the way of usual project processes
    during the election period. Technical decisions can be reversed by the
    incoming Steering Council if desired (and this would be evidence of the
    system working at intended). Governance changes require approval by the
    Board, and are thus unable to be snuck through inappropriately. Finally,
    the outgoing Steering Council is (as at all times) governed by the Django
    Code of Conduct, and so there is the continued check there. In short,
    there's not reason **not** to allow the normal processes to continue. 
```

shaib: Works beautifully for me, thanks.

tim-schilling: Works for me too. Thanks Carlton

final/0019-technical-governance-6x.rst
```diff
@@ -0,0 +1,760 @@
...
+* Allowing approving DEPs during elections.
+
+  * This is about not putting barriers in the way of usual project processes
+    during the election period. Technical decisions can be reversed by the
+    incoming Steering Council if desired (and this would be evidence of the
+    system working at intended). Governance changes require approval by the
+    Board, and are thus unable to be snuck through inappropriately. Finally,
+    the outgoing Steering Council is (as at all times) governed by the Django
+    Code of Conduct, and so there is the continued check there. In short,
+    there's not reason **not** to allow the normal processes to continue.
```
LilyFirefly:
```suggestion
    there's no reason **not** to allow the normal processes to continue. 
```

final/0019-technical-governance-6x.rst
tim-schilling: @carltongibson Did your change in https://github.com/django/deps/pull/107/commits/9e4783bfedfbefff9bff87288bfe374c139913cc address that here? The Security Team Charter discussion is still on-going. I'm not sure we need this to be open while that concludes though.

carltongibson: That was the intention. I'd assume a similar adjustment would help in the other discussion, but I didn't follow that so closely. Resolve away.

tim-schilling:
This was approved by the Steering Council. I'll merge this and then submit it to the DSF Board for them to accept as per DEP 10. If it's not accepted, we'll bring the status back to proposed and iterate on the necessary changes.

PR #109, opened 2026-04-24
Adopt Annual Release Cycle (DEP 20)

carltongibson:
Proposal to move Django to an Annual Release Cycle. 

As discussed: 

* https://buttondown.com/carlton/archive/an-annual-release-cycle-for-django/
* https://forum.djangoproject.com/t/should-we-adjust-djangos-versioning-to-use-a-form-of-calver/42811/

**Current status** I still need to work the motivation section fully, not yet ported from the introductory essay, but opening now to begin discussion.

**Update after initial reviews**, see https://github.com/django/deps/pull/109#issuecomment-4363591905

(Aside: given DEP 4 and DEP 44, I half wanted to number this DEP 444. )


draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+unresolved release blockers. The Steering Council may vote to delay a
+non-security release if it believes there is reason to do so.
+
+Versioning
+----------
+
+Django will use Calendar Versioning in the form ``YYYY.N``, where:
+
+* ``YYYY`` is the four-digit calendar year of the major release.
+* ``N`` is a monotonically incrementing counter within that year, starting
```
nanorepublica: I wonder if this could either start at `0` or be of the format `NNN`. The point here is not to create the assumption that this number represents the month in said year. Especially given that there is a monthly release. 

Unless of course we do want to closely match the month, but not quite if there is an urgent security release?

carltongibson: And also later in the cycle (years 2 and 3) many months may pass between releases. Making the second number the month was rejected (via the Forum discussion IIRC) because it would leave a question as to "Did I miss one" in such a case. 

`NN` might work. 

My preference is for `.1` over `.0` — it's the number of the release in the cycle. 

---

General point: Discussing this over the last few months, it's become clear that the _exact_ numbering scheme is a "What shade?" question (vs falling into line with the Python release process, and removing the gap between LTSs, which are the substance) 

As such, I'm happy to go with whatever emerges as consensus on the numbering scheme. The text presented here captures what I **think** was the balance of opinion so far. (I use YY for the year in my own projects, and initially favoured that, for example.)

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+  at ``1`` for the major release. Each subsequent bugfix release
+  increments ``N``.
+
+For example:
+
+* ``2028.1`` is the major release made in January 2028.
+* ``2028.2`` is the first bugfix release following ``2028.1``.
+* ``2029.1`` is the next major release, made in January 2029.
```
jefftriplett: This is amazing, btw! 

The "within that year" wording is ambiguous and may cause confusion. (Unless I'm confused by your intent.) 

Maybe it's worth spelling out or showing an example that even if one of those patches in the series lands in 2029, it might still remain in 2028.N series. 

(updated to switch from `.x` to `.N` to match the proposed change)

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+versions, replacing the eight-monthly cycle established in `DEP 4`__.
+
+__ https://github.com/django/deps/blob/main/final/0004-release-schedule.rst
+
+Under the proposed schedule, one major version of Django is released each
+calendar year, in January, and is supported for three years: one year of
+bugfix and security releases, followed by two years of security releases
+only. At any given time, three major versions are in support. The existing
+two-tier feature-release/LTS distinction is retired, with every release
+receiving what was previously the LTS-level commitment.
```
anze3db: As someone who's spent years warning clients that LTS releases offer a false sense of safety, I welcome this change! There was never a clear upgrade path / unified changelog from one LTS to the next, so migrating between LTS releases was always A LOT more stressful than upgrading to every single release in between.

Having no distinction between the LST/non-LTS releases makes things much clearer and easier!

draft/0020-annual-release-cycle.rst
benbacardi: Yes, this wasn't overly clear to me too. I presume a security fix released in 2029 would be applied to the 2028 release train as 2028.x?

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+
+Final major releases will be made in approximately the second week of
+January, carrying the new year's number. The exact release day remains at
+the Releaser's discretion, following the factors noted in DEP 44.
+
+Each major release is preceded by a pre-release phase:
+
+* An alpha release in early October, following the corresponding CPython
+  final release.
+* A beta release approximately four weeks before the release candidate.
```
RealOrangeOne: Question: This puts the beta release just as people are likely ramping down work for Christmas (other holidays also available). Is there concern that the beta will receive less testing because of that?

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+``YYYY.1``, bugfix support for ``(YYYY-1).1`` ends, and security support
+for ``(YYYY-3).1`` ends.
+
+Every release receives the same support commitment. The explicit "LTS"
+label used in DEP 4 is retired; see Rationale.
+
+Python Version Support
+----------------------
+
+Each Django major release supports three Python versions: the two Python
```
RealOrangeOne: Question: Do older Django versions receive support for newer Python versions? My reading is no, since supporting a newly-released Python version isn't a "bugfix [or] security release".

The downside of this being developers not only have to wait for 3 months before adopting newer Python versions, but they _also_ need to upgrade Django too.

accepted/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+  new cadence and numbering.
+
+* **Third-party package metadata.** ``Framework :: Django :: X.Y``
+  trove classifiers, tox environments, and CI matrices in the wider
+  ecosystem will need to be adjusted. Because the transition takes
+  effect at a clearly signposted release (``2028.1``), maintainers have
+  advance notice.
+
+No change is proposed to the support commitments already made for Django
+5.2 LTS or Django 6.2 LTS under the existing schedule.
```
RealOrangeOne, line 470: Thought: Depending on how long it would take to implement, should this be introduced for 6.2 instead? It means the "new world order" is fully migrated to earlier, resulting in less confusion in the interim

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+  trove classifiers, tox environments, and CI matrices in the wider
+  ecosystem will need to be adjusted. Because the transition takes
+  effect at a clearly signposted release (``2028.1``), maintainers have
+  advance notice.
+
+No change is proposed to the support commitments already made for Django
+5.2 LTS or Django 6.2 LTS under the existing schedule.
+
+References to, for example, Steering Council terms that are grounded in LTS
+cycles will need to be updated to simply state two-years, or as appropriate.
```
RealOrangeOne: Suggestion: Given LTS release are 3 years, it might make sense to align to that? It means everything follows a 3-year cycle. Granted it does increase the term substantially which may not be ideal.

RealOrangeOne (commented):
❤️

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+pre-release phase with the CPython release in October, giving Django a
+predictable window in which to validate against a newly released Python
+before Django's own final release in January.
+
+A January final release, carrying the new year's number, gives a
+"New Year, not Last Year" feel to each release and avoids the December
+holiday period for both the Releaser and users evaluating upgrades.
+
+Annual releases are (of course) slower than eight monthly ones, and that is a
+cost or this proposal. At this stage in Django's lifecycle, and given its
```
nessita: Typo?
```suggestion
cost for this proposal. At this stage in Django's lifecycle, and given its
```

draft/0020-annual-release-cycle.rst
carltongibson: I need to flesh this out more... 

The idea is that the current version picks up the new Python whilst it's within its first year, but the older two versions would not. 

There's no **requirement** to update your Python immediately to the newest version released, as the Django you're on will support 4, and the new one will only drop the last (that your current version supports) — If that makes sense. 

I'll spell this out more clearly.

accepted/0020-annual-release-cycle.rst
carltongibson: That seemed a bit quick to me 😬

draft/0020-annual-release-cycle.rst
carltongibson: Thanks, I'll clarify. 

The idea is to the December of the third year. (I said this better elsewhere...)

draft/0020-annual-release-cycle.rst
carltongibson: My personal take is that two years on the SC is just about right. 

But 🤷

(Either way, out of scope for this DEP — the note here is only that other docs will need to adjust phrasing.)

nessita (commented):
Thank you Carlton, I am very much in favor of all that is proposed here, with a minor concern on the versioning side (after some pondering).

I take your point that the numbering scheme is somewhat secondary to the substance, and I agree the substance is what matters. That said, I would rather we get the numbering right precisely because it is the part we cannot easily revisit later.

My concern is that CalVer feels like a one-way door. If the community later wants to return to a different cadence or scheme, the version numbers make that awkward in a way that 7.0, 8.0 style versioning simply does not. There is also a cross-year confusion I think is worth avoiding: a release of 2028.7 in March 2029 looks like a 2028 release but is arriving in 2029.

I would rather keep sequential versioning and bump the major once a year. No cross-year confusion (8.7 released in 2030 is unambiguously part of the 8.x series), no tooling changes, no one-way doors. And I think this also addresses most of the `Backwards Compatibility` concerns raised in the DEP: version string format stays the same shape, classifiers and tox environments need no format changes, and packaging tools keep working as-is. The only real migration work left is the LTS label and the schedule documentation, which we need to do regardless of the numbering scheme.

The only thing I think we lose is "at a glance you know the release year", but for that I propose:

* release 8.0 in January 2028 and increment by one per year
*  documented that year of release = major + 2020 (possible a one-liner in the docs and on the download page)

_(The skip of 7.x is a one-time cost and a natural signal to the ecosystem that something changed -- I much rather do math adding 2020 than 2019.)_

draft/0020-annual-release-cycle.rst
codingjoe: @carltongibson that would be different from Python, which starts patch releases with a `.0`. I know they don't use CalVer, but most CS engineers start counting at zero anyway.

“Skipping” the 0 might lead to confusion. I would certainly in my case.

Let's compare it to the biggest adopter of CalVar, that small 4 trillion dollar startup; they use iOS 26.0 (I hate the missing 20).

Lastly, 2028.1 would be the “first” patch for v2028. It feels easier to community care with a colleague in the office.

draft/0020-annual-release-cycle.rst
codingjoe: Hm… that's a lot of dinosaurs burned to test all versions. Especially since 3rd parties will have to support multiple Django versions. @tim-schilling has made a similar point previously. I'd like to hear his opinion too.

Personally, if a Python version doesn't receive bug fixes, any bugs like the garbage collection bug may need a backport in Django because an old Python version doesn't receive a patch.

People don't need to update neither Django nor Python for quite a while, wichout each Django version supporting a large Python version ragenes.

In short, I'd only support the 2 active Python versions at the time of release (without a hard pinned upper bound) so people can try a newer Python release.

carltongibson: @codingjoe OK... so the alternative there would be a _Green Only_ policy. (Just the current two Pythons at release) which brings it down another one. ("Another" because the current policy has the old LTS supporting five full versions at EOL.) 

I'm not 100% anti that. I went _Plus last yellow_ because that seemed to be the community vibe. 

But this is a topic we can decide on, and probably settle in a giant Zoom call 😅

carltongibson:
@nessita I don't dislike your suggestion. Thanks. My personal preference would be for CalVer because it carries additional semantic information (or carries it more clearly: you don't have to be an insider to understand when the EOL date is: Year + 3)

I think we should carry on this discussion and agenda it for the Giant Zoom Call™ (See reply to @codingjoe above/below  🥳)

draft/0020-annual-release-cycle.rst
jacobtylerwalls: We probably want a month between each:
- alpha: mid oct
- beta: mid nov
- rc: mid dec
- final: mid jan

So the beta will fall mid nov, not dec 1, and releasers can do the last RC before going "away" over holidays, and merge and backport any last minute bugs in early jan

Dropping the beta in mid nov seems fine, the flip side of work "ramping down" after november is that you have more "unexpected" time to play with the beta 😉

draft/0020-annual-release-cycle.rst
carltongibson: Yeah, again, for 0-based or 1-based I’m happy to go with the consensus. As I had it “1” is the first release. “2” is the second release and so on. (1-based is quite normal for CalVer schemes, I think. I certainly use them that way elsewhere.) 

I think beyond the first viewing either way would be clear enough.

accepted/0020-annual-release-cycle.rst
jacobtylerwalls: I think 7.0 works. If I count right, making 6.2 the first annual release would shorten its feature development period from 8 to 5 months this time. Better to roll this out with a more supple set of features, no? ("Django took a whole year and did what exactly?")

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+  new cadence and numbering.
+
+* **Third-party package metadata.** ``Framework :: Django :: X.Y``
+  trove classifiers, tox environments, and CI matrices in the wider
+  ecosystem will need to be adjusted. Because the transition takes
+  effect at a clearly signposted release (``2028.1``), maintainers have
+  advance notice.
+
+No change is proposed to the support commitments already made for Django
+5.2 LTS or Django 6.2 LTS under the existing schedule.
```
jacobtylerwalls: I think 7.0 works. If I count right, making 6.2 the first annual release would shorten its feature development period from 8 to 5 months this time. Better to roll this out with a more supple set of features, no? ("Django took a whole year and did what exactly?")

accepted/0020-annual-release-cycle.rst
nessita: @jacobtylerwalls so are you saying that on Jan 2027 we should do the final release of 7.0? (per my proposal of versioning, 2027.0 per this DEP).

I sort of like that, but we need to be fully aware that [202]7.0 will have effectively 4.5 months of feature development (from May 20th till Oct ~7th, when the new feature freeze would be in effect).

jacobtylerwalls: I was saying we should stick to the existing schedule for 6.2, which gives a feature freeze in January 2027, final release in April 2027. In January 2027 we start developing Django 7/2028, feature freeze October 2027, release January 2028.

If we don't accept this DEP, Django 7 would come out in December 2027. With this DEP, we move it by one month to January 2028. Very smooth 🙏

draft/0020-annual-release-cycle.rst
codingjoe: I can live with both, as I have been living with the current system. But in reality I believe there are two kinds of Django projects. The fossil that hasn't been updated since some agency wrote it and still runs on Django 1.8. And people who are on the prev. LTS but have an active ticket to upgrade.

Never have I experienced upgrading a Python version as a big consideration since Python 3. But mileage may vary.

codingjoe:
@carltongibson I have a question and maybe a suggestion:

I remember my first contributions and how daunting the feature freeze was. I was young, life was short, and the next release was an eternity away. With releases becoming an even rarer event, this might lead to more broken hearts. And it may also elevate the perception that Django is old (as in outdated), not my opinion, but I have JavaScript acquaintances…

What would it mean to **go from a feature freeze to a feature cutoff**? E.G., we create a 2028 branch at a certain cutoff day, but new features could be merged to main all year round. Patches get applied to all active version branches, including 2028.

My thought is this might prevent the awkward development if a feature misses the deadline (IMHO a real momentum killer).

I fear we are all looking at this with maintainer glasses, which is important, but the majority of people are Django users. And something like this could improve the narrative: Yes, one stable release a year, but we develop features all year round.

As a former fellow, or maybe also the current fellows: Is something like that even feasible, or am I daydreaming?

draft/0020-annual-release-cycle.rst
hugovk: Another +1 for starting at 0.

[pip](https://pypi.org/project/pip/#history) and [packaging](https://pypi.org/project/packaging/#history) come to mind as starting at 0. [attrs](https://pypi.org/project/attrs/#history) starts at 1.

.0 better aligns with the "major release" wording used elsewhere; "``2028.1`` is the major release" is a bit confusing at first, this sounds like a patch release.

(As an aside, I try and use "feature release" for CPython 3.x over "major release" to avoid no-we're-not-semver confusion.)

hugovk: Because there will be a 2028.13 in 2029 (or something like that), perhaps:

```suggestion
* ``N`` is a monotonically incrementing counter within that major release, starting
```

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+
+* ``YYYY`` is the four-digit calendar year of the major release.
+* ``N`` is a monotonically incrementing counter within that year, starting
+  at ``1`` for the major release. Each subsequent bugfix release
+  increments ``N``.
+
+For example:
+
+* ``2028.1`` is the major release made in January 2028.
+* ``2028.2`` is the first bugfix release following ``2028.1``.
```
hugovk: Perhaps include an example 2028.x that happens in 2029.

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+error).
+
+As per DEP 4, non-security releases may be held back if there are
+unresolved release blockers. The Steering Council may vote to delay a
+non-security release if it believes there is reason to do so.
+
+Versioning
+----------
+
+Django will use Calendar Versioning in the form ``YYYY.N``, where:
```
hugovk: Maybe link to https://calver.org/?

```suggestion
Django will use `Calendar Versioning <https://calver.org/>`__ in the form ``YYYY.N``, where:
```

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+tooling. The ``django-upgrade`` project provides fixers that rewrite user
+code across Django versions. Formalising any role for such tooling is out
+of scope for this DEP and is suitable material for a separate proposal.
+
+Backwards Compatibility
+=======================
+
+API stability commitments under this DEP are preserved: deprecated
+features continue to warn for at least two major releases before removal,
+and security support extends for three years from each release.
```
hugovk: Another plus: not only are stability commitments preserved, but the deprecation period is longer now, giving people more time to react.

accepted/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+The main compatibility impacts are on tooling and ecosystem expectations
+rather than on Django's Python API:
+
+* **Version string format.** ``YYYY.N`` is a new shape for Django's
+  version number. Code that parses or compares Django versions using
+  standard packaging tools (``packaging.version``, ``pip``'s resolver,
+  and similar) continues to work, since ``YYYY.N`` remains a
+  PEP 440-compliant, orderable version. Code that relies on ``X.Y``
+  having a bounded major component, or that uses ad hoc string parsing,
+  will need updating.
```
hugovk, line 445: Aside: this is a good opportunity to introduce tooling to identify code that makes assumptions about `django.VERSION` comparisons, similar to [Ruff’s YTT rules](https://docs.astral.sh/ruff/rules/#flake8-2020-ytt) and [Flake8’s flake8-2020 plugin](https://pypi.org/project/flake8-2020/).

shaib:
@codingjoe What you describe is, essentially, the way things have always worked. E.g. the `stable/6.0.x` branch was cut for the release of the first Alpha (2025-09-17 -- the second commit on that branch is tagged `6.0a1`), and development on `main` immediately continued towards 6.1.

shaib:
@nessita 
> 
> The only thing I think we lose is "at a glance you know the release year", but for that I propose:
> 
>     * release 8.0 in January 2028 and increment by one per year
> 
>     * documented that year of release = major + 2020 (possible a one-liner in the docs and on the download page)
> 
> 
> _(The skip of 7.x is a one-time cost and a natural signal to the ecosystem that something changed -- I much rather do math adding 2020 than 2019.)_

I like this. I would just comment that if we can skip one version number, we might as well skip 21, and get a lot of the benefits of your suggestion while keeping the year hint intact. That is, release 28.0 (or 28.1) in January 2028.

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+non-security release if it believes there is reason to do so.
+
+Versioning
+----------
+
+Django will use Calendar Versioning in the form ``YYYY.N``, where:
+
+* ``YYYY`` is the four-digit calendar year of the major release.
+* ``N`` is a monotonically incrementing counter within that year, starting
+  at ``1`` for the major release. Each subsequent bugfix release
```
sarahboyce: Given the quantity of security reports, it is likely we will have a release like `2028.36`, I think that's fine but just wanted to highlight it

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+Every release receives the same support commitment. The explicit "LTS"
+label used in DEP 4 is retired; see Rationale.
+
+Python Version Support
+----------------------
+
+Each Django major release supports three Python versions: the two Python
+versions with active upstream support at the start of the Django
+pre-release phase ("green"), plus the most recent Python version to have
+moved to security-only or end-of-life status ("plus last yellow").
```
sarahboyce: During the active support of bugfix and security releases for `2028.1` (actually should we just call this `2028` when referring to that release in general?) there will be a new Python released in the October of 2028 (Python 3.17), I assume we will backport support of this Python to `2028.1`?

accepted/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+
+* ``2028.1`` is the major release made in January 2028.
+* ``2028.2`` is the first bugfix release following ``2028.1``.
+* ``2029.1`` is the next major release, made in January 2029.
+
+Version numbers remain orderable under the standard comparison rules used
+by Python packaging tools. No third number is used; there is no separate
+"patch" component.
+
+Support Window
```
sarahboyce, line 102: This is maybe controversial but I am tempted that we end security support in the third Year after the October release. This aligns to when our oldest supported version of Python would drop security support.
As a compromise, I would release new versions of the pre-released Django with the security fixes (as `2029.a2` for example) so that folks can upgrade to a pre-release on production (if they wanted) knowing they will still get the security fixes.
This means that we are only ever managing 3 branches, all with supported versions of Python (so no back-porting Python security fixes)

draft/0020-annual-release-cycle.rst
sarahboyce: I also had a similar comment:
> If 2028.1 is released mid Jan, then 2028.rc1 (is that is what we will name it? or 2028.1.rc1) will be released 2 weeks before being roughly January 1st. I think that is logistically tough with Fellows vacation and likely vacation of translators. So perhaps 4 weeks before we do the release candidate, roughly mid December, 4 weeks before that the beta, roughly mid November, the alpha at the start of October

sarahboyce (commented):
I really like the idea of an annual cycle. I have no strong opinions of the naming. I have added some minor logistical comments

hugovk (commented):
+1 for annual release cycle. I also like CalVer (well, I am the author of [PEP 2026](https://peps.python.org/pep-2026/) :)

As a CPython release manager, the predictable annual release cycle makes things much easier. Also as a CPython developer, I know when the feature freeze and so on will happen, without having to always look things up.

Similarly as CPython user and library maintainer, it helps me plan: I know when to expect new and drop EOL versions. At Pillow, we have quarterly releases, and have aligned the October one to fall after the new CPython release.

accepted/0020-annual-release-cycle.rst
nessita: Makes sense, thanks Jacob! but the release version would not be 7.0, correct? it would be 2028.0 (or 28.0 or 8.0).

jacobtylerwalls: I'm unsure what the numbering scheme we will settle on, but yes, whatever number comes after 6.2, if that's what you're asking?

draft/0020-annual-release-cycle.rst
rik: Is there any reason to wait for the new Python final release to publish Django's alpha?

Python's RC1 is early August and RC2 is early September so it's already very representative of what will ship.

How about this calendar?
- Alpha: mid September (so Django's CI can run against Python RC1 in August and against RC2 before publishing it)
- Beta: mid October (after Python's final release)
- RC: mid November
- Final: early December (if two weeks between RC and final)

That way, Django is even closer to Python's schedule and everything is wrapped by the time of the Christmas/New Year holidays.

(With such a calendar, the version released in December 2027 would be called 2028)

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+As per DEP 4, non-security releases may be held back if there are
+unresolved release blockers. The Steering Council may vote to delay a
+non-security release if it believes there is reason to do so.
+
+Versioning
+----------
+
+Django will use Calendar Versioning in the form ``YYYY.N``, where:
+
+* ``YYYY`` is the four-digit calendar year of the major release.
```
rik: Starting a thread to gather the conversation on the major version format. I think of [@nessita's comment](https://github.com/django/deps/pull/109#pullrequestreview-4171628128) as "comment 0" of this thread.

I'm strongly in favour of using the year in the version number. I believe it greatly clarifies communication. It will probably help Django's users upgrade: it's easier to communicate to managers and other stakeholders that you need time to upgrade something called "Django 2023" than it is something called "Django 5".

I also believe it is great for Django's marketing. Many people not using Django believe it is old and does not receive meaningful updates. When those people read titles such as "Here's what's new in Django 2028", they get a more positive impression than reading "Here's what's new in Django 8".

rik (commented):
Very positive on the whole idea! Let's do this!

draft/0020-annual-release-cycle.rst
jacobtylerwalls: >  Is there any reason to wait for the new Python final release to publish Django's alpha?

I don't think there's a reason to wait for Python pre-releases, no. (We have a daily build against the in-development Python already.) Extrapolating from my experience last year, we'll get about a month's worth of "interesting" bugs relating to the new python into early November.

What I like about @rik's proposal:
- aligning even closer with the Python calendar. Under the "plus last yellow" plan, I think we still have Django support for an EOL python for about 3 months (from October to the end of the year). Your proposal shaves off a month there.

What I like less:
- dealing with a stream of release blockers over the holidays
- if we're going full "car versioning" instead of "cal versioning" by releasing "early", then let's just back up another month and release in November so that we can drop the last python as soon as it goes red at the end of October.

accepted/0020-annual-release-cycle.rst
jacobtylerwalls: > This is maybe controversial but I am tempted that we end security support in the third Year after the October release.

Or, we could achieve the same thing by just starting the three year cycles on November, a few weeks after the Python cycles start. Similar to @rik's idea above, but just backing up another month. (For example, release Django 2028.0-or-1 in Nov 2027.)

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,334 @@
...
+  annually since Python 3.9. Django's eight-month cycle is continually
+  misaligned with Python's, complicating Python support decisions and
+  increasing the matrix of supported interpreters.
+
+* **Removing the gap between LTS cycles.** Under the current policy, users
+  on an LTS release do not receive most bugfixes, which land only on the
+  current feature release. There are reports of users ceasing to file
+  issues with Django and instead maintaining private patches against
+  their LTS version. Making every release an LTS closes this gap: every
+  supported user receives bugfixes for a full year.
```
jacobtylerwalls: Glad to see this here, I'm happy about the bugfix window widening from 8 to 12 months. (I'm always sad about the regressions that just barely miss the shot clock.)

accepted/0020-annual-release-cycle.rst
sarahboyce: As a rough sketch of the updated calendar:
* Alpha: mid August (most GSOC projects finish in August, occasionally would tight to include but not a huge issue)
* Beta: mid September (usually around DjangoCon US but think that isn't a problem, beta is quite a simple release and talks on the upcoming release are possible)
* RC: mid October
* Final: early November

I like it :+1:

carltongibson: The idea with October alpha is that it picks up the newly released Python version at that point. 

Final in early November misses the whole "New Year, New Django" feel. (I feel this is more important than we're inclined to think it, as mere programmers: Not paying attention to how Django is perceived is a large part of the complaints we see in the ecosystem.) 

🤔

carltongibson: To the original point: 

> ... aligns to when our oldest supported version of Python would drop security support.

We could find some phrasing that says that we won't consider reports that affect only the EOL Python version for those last two months. 

(And, just to note, if we were to opt for the _Green Only_ policy, this concern wouldn't apply)

codingjoe: > Final in early November misses the whole "New Year, New Django" feel.

It's my birthday; I'd be OK 🤷‍♂️ 

Jokes aside, I agree with the sentiment to include "politics/perception" in our consideration as well as the community ergonomics.

A new Django version means updating Django support on community packages. Depending on your family, you have a lot of time during the holidays or none at all. Or you ditch your family for the annual CCC conference. 😏 

Considering that most companies will be stuck in let's hit that milestone before x-mas mode, I don't believe many will actually update before New Year's. That probably matters to Django, as adoption is a driver for regression detection. So ideally a release is situated to enable rapid community adoption.

carltongibson:
Just throwing a general comment here: thanks everybody for the constructive engagement. Keep it coming! — I will go through it, make clarifications and identify the open TBD topics probably over the weekend. (i.e. beginning of May, if you're reading from the future.)

accepted/0020-annual-release-cycle.rst
rik: > Final in early November misses the whole "New Year, New Django" feel. (I feel this is more important than we're inclined to think it, as mere programmers: Not paying attention to how Django is perceived is a large part of the complaints we see in the ecosystem.)

I agree that we need to pay attention to how it is perceived. As I said in my proposal, a Django released in early November would use the year of the upcoming year: in November 2027, we would release Django 2028.

One positive thing that can happen by having the new Django released close to the new Python is benefitting from combined news coverage.

draft/0020-annual-release-cycle.rst
rik: My example calendar was just illustrative. I'm happy if we find a sustainable way to ship closer to Python.

> * dealing with a stream of release blockers over the holidays

I'm not sure which holidays and which release you are referring to. @jacobtylerwalls Could you clarify?

jacobtylerwalls: I meant the regressions that come in during the first weeks after the final release. Given the relative lack of pre-release testing, this is the busiest period involving the release. Releasing in early December means this period will fall Dec 7-28 or so when we have less availability.

mjtamlyn:
Old "core dev" just chiming in here as I was more active when the current system (pre-LTS labelling) was designed...

I read the original article and completely agree. While I don't work full time in software any more and my last commit to Django was a long time ago, I do still have a number of hobby and work related sites, many of them over 10 years old, that run on Django. I can still seamlessly upgrade almost everything across many years of change and development, and I continue to be impressed by the API stability offered.

However as a casual user now I am completely out of the loop about how old various versions are, what happened when etc. Moving to a CalVer-like system will definitely help me to keep track of what's changed and when, and less frequent releases (12 months instead of 8) seems very reasonable for what is now an extremely stable piece of software.

I'm not gonna get involved in any discussion about the details, but this was basically a long way of saying... 👍

accepted/0020-annual-release-cycle.rst
jacobtylerwalls: > The idea with October alpha is that it picks up the newly released Python version at that point.

I still don't think we have to wait for that: someone who wanted to wait for the final Python release would just have an RC of Django to test with rather than an alpha, which is fine. If they want to test the Django alpha, they can just as easily use a Python RC with `3.xx-dev` in their matrix.

carltongibson:
I've pushed a round of clarifications based on the discussion so far. Where there were multiple views, I've tried to
lean into what seemed the dominant ones.

What's changed:

- .0 for the feature release; subsequent .1, .2, … are patch releases. "Feature release" used throughout. (@nanorepublica, @codingjoe, @hugovk, @sarahboyce)
- Dropped "within that year" from the N description; added a 2028.24 example to make late-cycle patches concrete. (@jefftriplett, @hugovk, @benbacardi)
- Restructured the Support Window into mainstream support (Year 1) and extended support (Years 2–3), with explicit definitions of what each tier backports and a December-transition paragraph. This goes a little beyond the literal review comments but makes the LTS-retirement story concrete.
- Pre-release timeline locked in: alpha early/mid October (post CPython final), beta early/mid November, RC early/mid December, final mid-January. (@jacobtylerwalls's monthly-spacing proposal, picked over the earlier-December variant on holiday/regression grounds — keeps the "New Year, New Django" framing.)
- Release Cadence now states explicitly that feature development is continual on main, and that branch-off at the alpha is the feature freeze. (@codingjoe, @shaib)
- Transition section now notes that Django 6.2 LTS continues to receive DEP-4 security support alongside the first annual releases, concluding April 2030. (@nessita, @jacobtylerwalls)
- New "Deprecation calendar" subsection in Rationale calling out @hugovk, @jacobtylerwalls's point: deprecation periods stretch from ~16 calendar months to ~24, without any change to the policy itself.

Still to decide:

- Versioning scheme — YYYY.N vs YY.N vs sequential majors (7, 8, 9…).

    On this, the lasting issue is that folks who aren't already _Inside Django_
    don't understand our versioning. The CalVer suggestion is to make it clear
    as day. When's it released: YYYY. When's it End of Life: YYYY+3. We can't
    do clearer than that. 

- Python support policy — "Plus Last Yellow" (three Pythons) vs "Green Only" (two).

We have a @django/steering-council and Fellow meeting coming up, so I reckon we can get at that then. 🥳

I still need to bring the Motivation points over from the original essay — that
section is still an outline. I think it's worth strengthening. 

Overall, I think we're getting there. 🚀

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,374 @@
...
+three feature releases are under support: the current release, under mainstream
+support, and the two prior years' releases, under extended support.
+
+Every release receives the same support commitment. The explicit "LTS"
+label used in DEP 4 is retired; see Rationale.
+
+Python Version Support
+----------------------
+
+At it's release, each Django feature release will support three Python
```
hugovk:
```suggestion
At its release, each Django feature release will support three Python
```

carltongibson:
Updating after the recent Steering Council and Fellows meeting. 

We have a broad consensus that (whilst in some ways appealing) the _Green Only_ Python support policy would be too aggressive, and so the DEP will focus on the _Plus Last Yellow_ policy, as written. 

That means each Django version will support the latest three Pythons at release, and pick up support for the newly released Python in the October of its first year. (Each Django version will thus support four Python versions in total.)

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,374 @@
...
+
+Version numbers remain orderable under the standard comparison rules used
+by Python packaging tools. There is no third numeric component; patch
+releases increment ``N``.
+
+Support Window
+--------------
+
+Every feature release is supported for three years from its date of
+release:
```
Stormheg: This part of the proposal excites me the most as a Django user and contractor!

Most of my clients stick with LTS releases of Django because there is not always enough time/money/willpower to upgrade Django in time when a non-LTS release goes out of support. Hence the preference to stick with LTS releases for long term continuity and security support.

The consequence is a bigger upgrade jump, like from v4.2 to v5.2, which this proposal would also solve. We could just upgrade to the next release to gain another year of support. In practice, we would probably upgrade to the latest release soon after it comes out in January.

If I imagine for a moment Django would have had this policy, I would have been able to use features from non-LTS releases like `GeneratedField` and the `django.tasks` framework much sooner.

my 2 cents from the perspective of a Django user 👍

Stormheg (approved):

codingjoe (approved):
<3 I love this a lot now <3

It makes maintaining my dependencies much simpler, and every project knows how many years it's behind. Can't wait!

In a few years, this will be so obviously "how you do it". That I want to halt and really appreciate the thoughtfulness from @carltongibson and everybody involved.

nanorepublica:
One thing I noticed with the new release of Python 3.15 coming soon is that Python has an Informational PEP detailing each release schedule (example of 3.16: https://peps.python.org/pep-0826/)

I wonder if there is any appetite us to adopt that as DEPs? I raise it here as I imagine this DEP would be the place to set that expectation going forward?

carltongibson:
@nanorepublica we used to do this on the wiki but the Fellows have moved this to a page of the site more recently:

https://www.djangoproject.com/download/6.1/roadmap/

I **suspect** that's probably sufficient, and likely has better visibility than the repo here. (Obviously we could promote DEPs better..) 

I think it's a separate proposal really.  Not necessarily a bad one...

hugovk (approved):
Following on from my earlier support -- https://github.com/django/deps/pull/109#pullrequestreview-4175407877 -- the recent changes are good and I support this DEP.

Especially the change around when alphas/beta/RCs are made (https://github.com/django/deps/pull/109/changes/45fb759ce569826a4e1f1537f80a0baa50b34fe6), for example, "A release candidate approximately two weeks before the final release" -> "A release candidate in early/mid December".

This gives the Releaser more flexibility to manage schedules, in particular to ensure they can avoid making releases around end-of-year holidays and that they can have a proper break.

carltongibson:
OK, thanks all. I've pushed a series of changes addressing the various points raised and remaining TODOs. 

I was going to resolve all the conversations but reviewing them I don't want to do that, as I think there's valuable points raised by a varied range of folks. 

I'll be asking the @django/steering-council council to progress this DEP. Please do keep any input coming. 

I've been humbled by how constructive the whole process has been, so thank you all again. 🎁

accepted/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,494 @@
...
+(EOL) two months before the corresponding Django version. At this point, Django
+will continue to test against the then EOL Python, in order to avoid
+regressions in line with the stability policy. Nonetheless, Django may decline
+to address an issue affecting only the then EOL Python version, if such emerges
+in the final two months of its support window.
+
+Deprecation Policy
+------------------
+
+If a feature is deprecated in release ``A``, a deprecation warning is
```
LilyFirefly, line 161: Nitpick: I stumbled slightly when reading this because it says `A, a`. I think using a different letter would be more readable, perhaps `X`?

LilyFirefly (approved):
While I'm a bit sad that the release cycle is slowing down, I think the pros of this proposal outweigh that one con.

accepted/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,494 @@
...
+raised in ``A`` and ``A+1``, and the feature is removed in ``A+2``. This
+preserves the existing guarantee that a deprecated feature remains
+available, with warnings, across at least two feature releases.
+
+Because every release is supported for three years, a user on any supported
+version has at least two feature releases in which to respond to any
+deprecation before it is removed.
+
+Transition
```
sarahboyce, line 170: Potentially what the supported versions image will look like assuming I have understood the DEP correctly
Image

Note that I started making very very rough changes to djangoproject.com to see what this might look like: https://github.com/django/djangoproject.com/pull/2733

draft/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,494 @@
...
+The current two-tier system has a well-documented failure mode: the
+backport policy means that users on an LTS receive only a small subset of
+fixes, while users who stay current receive bugfixes but face more
+frequent upgrades. Collapsing this into a single tier — every release
+supported for three years — removes the failure mode without reducing the
+commitment offered to any user.
+
+The explicit "LTS" label is retired in this DEP on the grounds that every
+release now carries the commitment that label previously denoted. It
+remains the responsibility of the Marketing Working Group, the
```
tim-schilling: The marketing working group isn't a formal WG and should probably be replaced by the SC and / or board.

There's another bit in the rationale that needs updating.

tim-schilling (approved):

accepted/0020-annual-release-cycle.rst
```diff
@@ -0,0 +1,494 @@
...
+
+References to, for example, Steering Council terms that are grounded in LTS
+cycles will need to be updated to simply state two-years, or as appropriate,
+but such edits are out of scope for this DEP.
+
+Reference Implementation
+========================
+
+There is no reference implementation at the time of writing. Adoption of
+this DEP requires changes including:
```
tim-schilling, line 480: Not sure if I missed this part of the discussion, but depending on when we implement this versioning scheme, we should identify how we're managing that with other deprecation errors/warnings. People likely have those coded into their applications, so we should consider maintaining those for backwards compatibility.

nanorepublica:
This person is clearly frustrated this morning (https://fosstodon.org/@mgorny@treehouse.systems/117041743046159003)

But might be worth engaging them on this DEP to see if it's an improvement for them? I think there is a legitimate crossover of topics?

sarahboyce:
> This person is clearly frustrated this morning (https://fosstodon.org/@mgorny@treehouse.systems/117041743046159003)
> 
> But might be worth engaging them on this DEP to see if it's an improvement for them? I think there is a legitimate crossover of topics?

I think they are backporting support for Python 3.15 into every supported version of Django (including 5.2 and 6.0 which is only getting security fixes) when Python 3.15 final isn't even released yet. I don't think these changes will help them

accepted/0020-annual-release-cycle.rst
carltongibson: 67b53a9426c4099540ae6876fecf03c2533b3806

LilyFirefly: Ah, there were (at least) two places using `A` and you only fixed one of them.

draft/0020-annual-release-cycle.rst
carltongibson: f69fe337e5d0653013b3a9f43e9834836b76d42b

accepted/0020-annual-release-cycle.rst
carltongibson: 27c419bcac1af5a753f51b171c8a6bed876b08b1

nanorepublica:
Thanks for the clarification  @sarahboyce !

carltongibson:
Thank you again for the constructive discussion all. 🎁

DEP 20 was accepted by the Steering Council in their meeting of 2026-08-03.

As such, I've moved this to the correct folder, and will squash and merge this PR. 

I'll quote two paragraphs from the Transition section here: 

> The first release under this DEP is Django ``2028.0``, replacing what
would otherwise have been Django 7.0 (scheduled under the existing cycle
for December 2027). ``2028.0`` is released in January 2028 and is
supported for three years on the terms above.
> 
> Django 6.2 LTS, released April 2027, continues to receive security
support per DEP 4 alongside the first feature releases under this DEP,
with its support window concluding in April 2030.

As per the Reference Implementation section: 

> * Communicating the change via the Django blog and other marketing efforts.

We will follow up with a summary post on the Django blog in the next couple of weeks. That will likely be the best tl;dr place to link folks to.

Let's go! 🕺

PR #105, opened 2026-02-09
DEP 0018 Dictionary-based EMAIL_PROVIDERS (ticket-35514)

medmunds:
📄 Read the formatted [draft DEP 0018, latest revision](https://github.com/medmunds/deps/blob/email-providers/draft/email-providers.md).

See [ticket-35514#comment:24](https://code.djangoproject.com/ticket/35514#comment:24)

Django ticket-35514 will implement a dictionary-based `EMAIL_PROVIDERS` setting, similar to `CACHES`, `DATABASES`, `STORAGES` and `TASKS`. The ticket was approved following discussion on the django-developers list and at DjangoCon Europe 2024 sprints.

The purpose of this DEP is to facilitate discussion and decisions on the proposed API and related deprecations.

PR #111, opened 2026-05-14
DEP 0018 Mailers cleanup (f.k.a. email providers)

medmunds:
Followup from PR #105 and some other cleanup:
* Rename `EMAIL_PROVIDERS` to `MAILERS` (assisted by GPT-5.5)
* Clarify deprecation warning when no email-related settings are defined
* Update PR and status
* Add `sendtestmail --using ALIAS` option as suggested future work

<del>Also, I'd suggest renaming the file to `0018-mailers.md` at some point—perhaps when it it is moved out of the draft directory. (Renaming the file will likely complicate reviewing any other changes in GitHub.)</del> Edit: renamed and moved. (Looks like GitHub's compare view figured it out.)

### DIFF
--- a/accepted/0018-mailers.md
+++ b/accepted/0018-mailers.md
@@ -3,28 +3,28 @@ DEP: 0018
 Author: Mike Edmunds
 Implementation Team: Jacob Rief, Mike Edmunds
 Shepherd: Natalia Bidart
-Status: Draft
+Status: Accepted
 Type: Feature
 Created: 2026‑02‑09
-Last-Modified: 2026‑04‑21
+Last-Modified: 2026‑05‑20
 ---
-# DEP 0018: Dictionary-based EMAIL_PROVIDERS settings
+# DEP 0018: Dictionary-based MAILERS setting and mailers factory
 
 **Table of Contents**
 
 - [Abstract](#abstract)
 - [Motivation](#motivation)
 - [Specification](#specification)
-  - [`EMAIL_PROVIDERS` setting](#email_providers-setting)
+  - [`MAILERS` setting](#mailers-setting)
   - [New exceptions](#new-exceptions)
   - [`using` argument to send functions](#using-argument-to-send-functions)
-  - [`providers` factory](#providers-factory)
+  - [`mailers` factory](#mailers-factory)
   - [Updates to built-in EmailBackend classes](#updates-to-built-in-emailbackend-classes)
   - [Related updates to other Django code](#related-updates-to-other-django-code)
 - [Backwards compatibility](#backwards-compatibility)
   - [Deprecated email settings](#deprecated-email-settings)
   - [Settings compatibility](#settings-compatibility)
-  - [Default provider compatibility](#default-provider-compatibility)
+  - [Default mailer compatibility](#default-mailer-compatibility)
   - [`using` arg compatibility](#using-arg-compatibility)
   - [Testing outbox compatibility](#testing-outbox-compatibility)
   - [`fail_silently` sending option deprecated](#fail_silently-sending-option-deprecated)
@@ -35,12 +35,14 @@ Last-Modified: 2026‑04‑21
   - [Third-party compatibility](#third-party-compatibility)
   - [Upgrading EmailBackend implementations](#upgrading-emailbackend-implementations)
   - [django-upgrade recommendations](#django-upgrade-recommendations)
+- [Naming](#naming)
 - [Reference implementation](#reference-implementation)
 - [Future work](#future-work)
+  - [Future: sendtestemail using option](#future-sendtestemail-using-option)
   - [Future: System checks](#future-system-checks)
-  - [Future: Password reset email provider](#future-password-reset-email-provider)
-  - [Future: Provider-specific message defaults](#future-provider-specific-message-defaults)
-  - [Future: Cached `providers`](#future-cached-providers)
+  - [Future: Password reset mailer](#future-password-reset-mailer)
+  - [Future: Mailer-specific message defaults](#future-mailer-specific-message-defaults)
+  - [Future: Cached `mailers`](#future-cached-mailers)
 - [Prior art](#prior-art)
 - [History](#history)
 - [AI disclosure](#ai-disclosure)
@@ -49,9 +51,9 @@ Last-Modified: 2026‑04‑21
 
 ## Abstract
 
-This DEP proposes supporting multiple email provider configurations, via:
-* a new dictionary-based `EMAIL_PROVIDERS` setting
-* a new `mail.providers[alias]` factory
+This DEP proposes supporting multiple email configurations, via:
+* a new dictionary-based `MAILERS` setting
+* a new `mail.mailers[alias]` dict-like factory
 * adding `using=alias` args to email sending functions
 
 This will align Django's email backend configuration with similar capabilities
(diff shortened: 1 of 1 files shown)


accepted/0018-mailers.md
codingjoe, line 1: You might want to update the filename too.

LilyFirefly, line 1: This should include the DEP number: `draft/0018-email-providers.md`.
It might also make sense to rename to `draft/0018-mailers.md`.
Finally, perhaps we should move this to `accepted/0018-mailers.md` in this PR?

Edit - oh, I see you mentioned this in your PR summary!

draft/email-providers.md
```diff
@@ -6,25 +6,25 @@ Shepherd: Natalia Bidart
 Status: Draft
```
LilyFirefly: Should we move this to `Accepted`?

codingjoe (commented):
Hm… now that I see it, I have become hesitant. It reads funny (but English is my second language).

Naming things is hard, but if I have learned anything in the past years, inventing words never pans out.

So I read into the [RFC 5321](https://www.rfc-editor.org/rfc/rfc5321.html#section-2.3.10) and here they are explicitly referred to as **Gateway**.

> A "gateway" SMTP system (usually referred to just as a "gateway")
   receives mail from a client system in one transport environment and
   transmits it to a server system in another transport environment.
   Differences in protocols or message semantics between the transport
   environments on either side of a gateway may require that the gateway
   system perform transformations to the message that are not permitted
   to SMTP relay systems.  For the purposes of this specification,
   firewalls that rewrite addresses should be considered as gateways,
   even if SMTP is used on both sides of them (see [RFC 2979](https://www.rfc-editor.org/rfc/rfc2979) [[27](https://www.rfc-editor.org/rfc/rfc5321.html#ref-27)]).

We are currently only supporting RFC 5321 (SMTP) messages, but let's look into the future: SMS calls them gateways too. RCS is only a message spec, the SIP (RFC 3261) for transport. Here gateway refers to PTSN bridges. [But no sane person will ever build their own VoIP stack.](https://github.com/codingjoe/VoIP)

---

TL;DR: My conclusion: `EMAIL_GATEWAYS` and `django.core.mail.gateways` and `django.core.mail.default_gateway`.

carltongibson:
Errr... :)

So we have `DATABASES` and `STORAGES`, so I think I would have gone for `EMAIL_PROVIDERS` but looking at the diff, that was already rejected, yes? And I don't know why, so I can't really comment. 

TBH I'm not sure about "gateways": it seems a bit removed from what we're doing, and it's the bit at the other end of the pipe no? (Not sure.) 

`EMAIL_BACKENDS`? Too close to the old setting? 

The motivation was to be able to select multiple email _providers_, each using a _backend_ with a given _config_... — Any of those nouns in plural might make sense. 

_Mailers_? Maybe, yeah. But I don't think I've ever said _"mailer"_ before, so I do see the concern.

Sorry, that's not particularly definitive.

LilyFirefly:
@carltongibson Previous discussion started here: https://github.com/django/deps/pull/105#pullrequestreview-4227011246

I still have a soft spot for `EMAILS`, but I can live with any of the other options.

carltongibson:
Thanks @LilyFirefly! 

It seems like `EMAIL_PROVIDERS` had a good amount of support there. 

🤷

codingjoe:
I believe it was me who re-raised concerns about `EMAIL_PROVIDERS`. 🙈
While those concerns may hold some value, I certainly prefer it over a synthetic name.

I can confidently say I have used the term _email provider_ in a conversation.
Meanwhile, apparently, at least in the PHP community ([Symfony](https://symfony.com/doc/current/mailer.html), [Laravel](https://laravel.com/docs/13.x/mail)) uses _mailer_ as a term.

> I still have a soft spot for EMAILS, but I can live with any of the other options.

I really like it for its simplicity too, and it's consistent with `DATABASES` & Co.

ryanhiebert:
I have a soft spot for variables to be as short as would be unambiguous. EMAILS and MAILERS both fit the short part for me. EMAILS feels possibly ambiguous to me, because a reasonable intuition is a collection of email addresses, rather than email providers, so MAILERS is my preference.

Some have contested that "mailers" is an unknown term, and while I don't think it's the most common way to talk about a relay, I have seen that and it feels intuitive to me, especially since the Django package name is `django.core.mail`. For example, I have for many years used the `nullmailer` program as a thin sendmail replacement to relay local messages on my Debian servers to an SMTP server.

jacobtylerwalls:
It's a real-world word, too ([photo mailers](https://www.amazon.com/photo-mailers/s?k=photo+mailers)). The Symfony and Laravel precedents are reassuring.

cc/ @jrief

tim-schilling:
In order of preference:

1. emailers
2. email_providers / email_<backends> (or any other noun)

There's a gap here separating the above two from the below for me.

3. mailers (see below)
4. emails (likely to conflict with a project setting)
5. gateways (see below)

I'm honestly less worried about conforming to an RFC's naming scheme and more worried about what new devs are going to expect to look for in our configuration to tweak this. I wouldn't ever look for a gateway. I think it'd be reasonable to search our settings for "email" which would mean "mailers" and "gateways" wouldn't turn up.

nessita:
The rename to `MAILERS` was done following the discussion in PR #105, where the we reached what I considered consensus. That rename was a significant amount of work for Mike and myself, so unless there is a very clear, concrete gain, I am firmly against reopening this (and I don't see that gain in the current discussion). On the merits for `MAILERS`:

* It has precedents: "Mailer" is not a synthetic term, it has a [true English meaning](https://www.merriam-webster.com/dictionary/mailer), and it's used with the same semantic in `send_mail`, `send_mass_mail`. As Jacob said there are real-world usages (besides the mentioned packages, there is also [nullmailer](https://wiki.debian.org/nullmailer)).

* It follows Django's established convention of plural resource settings using short, unqualified nouns: `DATABASES`, `CACHES`, `STORAGES`, `TASKS`. The corresponding factory objects follow the same pattern: `connections`, `caches`, `storages`. `EMAIL_PROVIDERS` breaks it: no other analogous setting uses a compound `NOUN_VERB` form. The Django module is already named `django.core.mail`, so `mail.mailers` reads as naturally as `cache.caches` or `storage.storages`.

* Regarding the alternatives: `EMAILS` reads as a collection of message objects, not sending configurations. As said `EMAIL_GATEWAYS` and `EMAIL_RELAYS` are not suitable. `EMAILERS` has no precedent and `EMAIL_PROVIDERS` carries the connotation of a cloud vendor, not a backend configuration.

* On discoverability: the argument weakens when considering that `connections` won't appear when searching for "database", and `storages` won't appear when searching for "file".

The name is good and accurate, and I'd kindly ask everyone to consider the cost of continued indecision against the cost of another rename.

jrief:
For me, `MAILERS` sounds perfectly reasonable.

@nessita why didn't we come up with that name in Vigo in 2024? Didn't we even consider it?

nessita:
> For me, `MAILERS` sounds perfectly reasonable.
> 
> @nessita why didn't we come up with that name in Vigo in 2024? Didn't we even consider it?

We did not consider it, no one suggested it. When Mike mentioned it recently in the other PR, it seemed ideal.

nessita:
@django/steering-council from you view, is this DEP ready to be merged as accepted?

frankwiles (approved):
LGTM! 🚀

tim-schilling:
@nessita the steering council is in agreement with this change. :heavy_check_mark: Thank you and @medmunds for all your work!

nessita:
@medmunds I would like to merge this, but before doing so it may be a god idea to tidy up the commits a bit. I'm thinking we want want to reference the DEP somehow, and evaluate if we can squash down to 1. Would you have availability to do that?

medmunds:
@nessita I've squashed the commits and updated the commit message. (No changes to the content.)

PR #112, opened 2026-05-21
Prototype static sphinx website for the DEPs

nanorepublica:
I had Claude build this one morning. I think it works as a starting point, which I presented as a solution in this Forum thread: https://forum.djangoproject.com/t/moving-files-in-dsf-repos/45034

You can see an example output site here: https://tranquil-pavlova-402378.netlify.app

I'm opening this as I think it's a meaningful output improvement, but I can understand if it's closed as Claude has clearly written this. I have read it and it seems reasonable to me, but I'm open to other solutions. 

Long term this could be deprecated and something built into the main project website.

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -1,0 +1,46 @@
+name: Build & deploy DEPs site
+
+on:
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: pages
+  cancel-in-progress: false
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+          cache: pip
+          cache-dependency-path: tools/requirements.txt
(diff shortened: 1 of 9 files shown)


codingjoe:
Ref [ticket-37099](https://code.djangoproject.com/ticket/37099)

codingjoe:
@nanorepublica cool idea, but can we integrate this into the [djangoproject.com Sphinx build](https://github.com/django/djangoproject.com)? I am hesitant to add more domains. Having everything in one place sure is nice from a user perspective.

nanorepublica:
@codingjoe when you say integrate, do you mean the Sphinx build for Django that gets published to the website at docs.djangoproject.com ? 

I think there are also some typo fixes to make all the DEPs more consistent to ease the generation of the website. I'll pull those out for separate PRs asap.

codingjoe:
@nanorepublica Yes, but now that you mention it, it's already on `docs.`, so maybe a `deps.` doesn't hurt? Hard to say; I find the range of subdomains with forums, wikis, and co. already pretty confusing.

nanorepublica:
doing `deps` would be much simpler - it would keep this site in terms of infrastructure, then adding a custom domain to it.

Noted on the number of subdomains... perhaps something the OCWG can look in due course.

PR #106, opened 2026-03-01
Added draft DEP for ORM support for Common Table Expressions (CTE)

Genarito:
Hi!

Here is a DEP describing the proposed integration of Common Table Expressions support into the Django ORM core. This document was requested by @nessita in Django issue [#20713](https://github.com/django/django/pull/20713).
I would be happy to receive any feedback you may have.

Best regards!
Genaro

PR #113, opened 2026-06-01
Add  support for table value expressions in ORM

p-r-a-v-i-n:
This DEP proposes the implementation of a CompositeField within the Django ORM to support Table-Valued Expressions (TVEs) and functions that return sets of rows.

Refereneces: 
- Ticket: https://code.djangoproject.com/ticket/37115
- Django New Features Thread: https://github.com/django/new-features/issues/25
- forum posts:
        - https://forum.djangoproject.com/t/proposal-add-generate-series-support-to-contrib-postgres/21947/4

### DIFF
--- a/draft/table_value_expression.rst
+++ b/draft/table_value_expression.rst
@@ -145,6 +145,41 @@ There are three aspects to this that we'd like to work on:
 Specification
 =============
 
+Composite output metadata
+-------------------------
+
+The first requirement is a way for an expression or query source to describe
+multiple output columns.
+
+For scalar expressions, Django already uses ``Expression.output_field``. For
+example, a subquery returning only ``subject`` can expose a single field such as
+``CharField``.
+
+For table expressions that expose more than one column, Django needs an
+equivalent representation for multiple named fields. For example:
+
+.. code-block:: python
+
+   JsonEach("payload")
+   # exposes:
+   #   key -> TextField()
+   #   value -> JSONField()
+
+This may be implemented as a lightweight composite output field, or as part of
+a more general ``CompositeField`` design. The initial scope should be limited to
+describing expression/query output, not full model-field behavior.
+
+This matters because the ORM must preserve Django's existing lookup behavior.
+If ``item__key`` resolves to a ``TextField``, then normal text lookups and
+transforms should continue to work.
+
+The exact public API is still open.
+
+Expression-like table sources
+-----------------------------
+
+After output columns can be represented, the ORM needs a way to register an
+expression or query as a table source in ``FROM`` or ``JOIN``.
 
 This DEP uses ``TableExpression`` as a placeholder name. The final public API
 may use a different name.
@@ -184,9 +219,9 @@ For a multi-column source:
 The implementation should preserve Django's existing lookup behavior by
 associating each exposed column with a Django field.
 
-The exact mechanism is open. One possible direction is to represent
-multi-column output through ``Expression.output_field`` using a composite output
-representation.
+The exact mechanism is open. One possible direction is to represent this through
+``Expression.output_field`` using the lightweight composite output metadata
+described above.
 
 Table Source Compilation
 ------------------------
@@ -264,38 +299,6 @@ reusable scalar expressions.
 Alternatives Considered
 =======================
 
-Composite Output Metadata
--------------------------
-
-Another approach is to represent the output shape through ``output_field``:
-
-.. code-block:: python
-
-   class JsonEach(Func):
-       function = "json_each"
-       output_field = CompositetField([
-           ("key", models.TextField()),
-           ("value", models.JSONField()),
-       ])
-
-This may integrate better with the rest of the ORM, but it requires more
-exploration of Django's field and expression internals.
-
-Relation-Style API
-------------------
-
-One possible design is a dedicated relation wrapper:
-
-.. code-block:: python
-
-   Sales.objects.alias(
-       item=Relation(JsonEach("payload"))
-   )
-
-That design may require helper objects to bridge relation aliases to column
-references.
-
-
 Explicit Column Metadata
 ------------------------
 
@@ -314,6 +317,7 @@ An early implementation could accept column metadata directly:
 This may be easy to prototype, but it creates a separate metadata path from
 ``Expression.output_field``.
 
+You can find more detail `here <https://github.com/p-r-a-v-i-n/Generic---Relation---API-Design/blob/main/RELATION_API_BLUEPRINT.md>`_.
 
 Copyright
 =========
@@ -321,4 +325,3 @@ Copyright
 This document has been placed in the public domain per the Creative Commons
 CC0 1.0 Universal license
 (http://creativecommons.org/publicdomain/zero/1.0/deed).
-
--- a/draft/table_value_expression.rst
+++ b/draft/table_value_expression.rst
@@ -53,7 +53,7 @@ But this breaks when we are dealing with multi-column results. For example:
 
    Sales.objects.annotate(
        first_subject=first_object
-   ).values("first_subject")
+   ).values("first_subject__subject", "first_object__duration")
 
 This will raise `Cannot resolve expression type, unknown output_field`. because the ORM does not have an expression representing multiple columns and thus cannot make use of a join against the multi-column subquery.
 
@@ -67,6 +67,14 @@ used as a scalar annotation, Django renders it in the ``SELECT`` list:
 .. code-block:: python
 
    Sales.objects.annotate(series=GenerateSeriesFunc(1, 3))
+
+
+Expected SQL:
+.. code-block:: sql
+
+   SELECT "SRF_sales"."id", "SRF_sales"."sold_on", "SRF_sales"."day", "series"."value" AS "series"
+   FROM "SRF_sales"
+   CROSS JOIN LATERAL generate_series(1, 3) AS "series"("value")
 
 The actual SQL generated:
 
@@ -87,7 +95,15 @@ expression:
 
 .. code-block:: python
 
-   Sales.objects.annotate(series=GenerateSeriesFunc(1, 3)).filter(series__gt=1)
+   Sales.objects.alias(series=GenerateSeriesFunc(1, 3)).filter(series__gt=1)
+
+Expected SQL:
+.. code-block:: sql
+
+   SELECT "SRF_sales"."id", "SRF_sales"."sold_on", "SRF_sales"."day", "series"."value" AS "series"
+   FROM "SRF_sales"
+   CROSS JOIN LATERAL generate_series(1, 3) AS "series"("value")
+   WHERE "series"."value" > 1
 
 The actual SQL generated:
 
@@ -143,9 +159,7 @@ expression, it could expose named columns:
 .. code-block:: python
 
    Sales.objects.alias(
-       cohort_data=TableExpression(
-           Cohort.objects.values("subject", "duration")[:1]
-       )
+       cohort_data=Cohort.objects.values("subject", "duration")[:1]
    ).values("cohort_data__subject", "cohort_data__duration")
 
 
@@ -167,76 +181,53 @@ For scalar expressions, Django already uses ``Expression.output_field``. For
 example, a subquery returning only ``subject`` can expose a single field such as
 ``CharField``.
 
-For table expressions that expose more than one column, Django needs an
-equivalent representation for multiple named fields. For example:
+For expressions that expose more than one column, Django needs an
+equivalent representation for multiple named fields. For example, a subquery or table expression might expose:
 
 .. code-block:: python
 
-   JsonEach("payload")
    # exposes:
-   #   key -> TextField()
-   #   value -> JSONField()
+   #   title -> CharField()
+   #   body -> TextField()
 
-This may be implemented as a lightweight composite output field, or as part of
-a more general ``CompositeField`` design. The initial scope should be limited to
-describing expression/query output, not full model-field behavior.
+To solve this, this proposal introduces a ``CompositeField``. This acts as a lightweight composite output field designed
+specifically for describing multi-column expression/query output, rather than acting as a concrete database model field.
+This distinction matters because the ORM must preserve Django's existing lookup behavior.
+By wrapping the output in a ``CompositeField``, if ``posts__title`` resolves to a ``CharField``, then all normal
+text lookups and transforms (like ``__icontains``) will automatically continue to work as expected.
 
-This matters because the ORM must preserve Django's existing lookup behavior.
-If ``item__key`` resolves to a ``TextField``, then normal text lookups and
-transforms should continue to work.
-
-The exact public API is still open.
-
-Expression-like table sources
------------------------------
-
-After output columns can be represented, the ORM needs a way to register an
-expression or query as a table source in the ``FROM`` or ``JOIN``.
-
-We will take inspiration from the existing `FilteredRelation`. We would follow the
-mechanism it used to register itself as a table source in the ``FROM`` clause.
-
-This DEP uses ``TableExpression`` as a placeholder name. The final public API
-may use a different name.
-
-Example:
-
-.. code-block:: python
-
-   Sales.objects.alias(
-       series=TableExpression(GenerateSeries(1, 3))
-   ).filter(series__value__gt=1)
-
-For single-column table expressions, the final API may allow a shorter form such
-as ``series__gt=1``. That detail is still open.
-
-Handling multi-columns
-----------------------
+Handling multi-columns single row
+---------------------------------
 
 A table expression must describe the columns it exposes.
-
-For a single-column source:
-
-.. code-block:: python
-
-   GenerateSeries(1, 3)
-   # exposes: value -> IntegerField()
 
 For a multi-column source:
 
 .. code-block:: python
 
-   JsonEach("payload")
-   # exposes:
-   #   key -> TextField()
-   #   value -> JSONField()
+   subquery = Post.objects.filter(user=OuterRef("pk")).values("title", "body")[:1]
+   User.objects.alias(posts=subquery).values("posts__title", "posts__body")
 
-The implementation should preserve Django's existing lookup behavior by
-associating each exposed column with a Django field.
+On the fly we will set the ``output_field`` of the subquery to CompositeField.
+This could be achieved when the initial ``resolve_ref()`` is triggered to get the transform.
+When Django tries to extract the subfield, it delegates to ``BaseExpression.get_transform()`` which looks up ``self.output_field``.
 
-The exact mechanism is open. One possible direction is to represent this through
-``Expression.output_field`` using the lightweight composite output metadata
-described above.
+This seamlessly triggers ``sql/Query.output_field`` on the inner query, which is exactly where we can evaluate
+the query's select list and dynamically form the ``CompositeField`` to handle the extraction.
+
+Expression-like table sources
+-----------------------------
+After output columns can be represented, the ORM needs a way to explicitly register an expression or query as a table source in the ``FROM`` clause.
+We can be sure if subqueries return multi-column single row or multi-column multi-row. For example:
+.. code-block:: python
+   subquery = Post.objects.filter(user=first_user).values("title", "body")
+   User.objects.alias(posts=subquery).values("posts__title", "posts__body")
+In this example, the ORM cannot reliably know if the ``subquery`` returns exactly one row or multiple rows.
+To solve this, our approach requires the developer to explicitly declare when a result is a multi-row table source. The user will do this by wrapping the queryset in a ``TableExpression``.
+.. code-block:: python
+   User.objects.alias(posts=TableExpression(subquery))
+This explicit wrapper allows the ORM to cleanly separate the processing logic. When encountering a ``TableExpression``, the ORM knows to place the subquery inside the ``FROM`` clause (utilizing ``LATERAL`` joins if there are outer references) rather than resolving it as a scalar subquery in the ``SELECT`` clause.
+*(Note: This DEP uses ``TableExpression`` as a placeholder name. The final public API may use a different name).*
 
 Table Source Compilation
 ------------------------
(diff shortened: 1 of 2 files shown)


draft/table_value_expression.rst
```diff
@@ -0,0 +1,329 @@
...
+.. code-block:: python
+
+   JsonEach("payload")
+   # exposes:
+   #   key -> TextField()
+   #   value -> JSONField()
+
+This may be implemented as a lightweight composite output field, or as part of
+a more general ``CompositeField`` design. The initial scope should be limited to
+describing expression/query output, not full model-field behavior.
```
p-r-a-v-i-n: We planned that the design of  `CompositeField` would work for  `virtual sources(expressions)` first.

draft/table_value_expression.rst
```diff
@@ -0,0 +1,329 @@
...
+   ).values("first_subject")
+
+The reasons this works:
+
+* ``.values("first_subject")`` is treated like a selected column of
+  the ``Sales`` query.
+* This happens because Django renders the annotation in the ``SELECT`` clause.
+  That is valid for scalar expressions.
+
+But this breaks when we are dealing with multi-columns results. for example:
```
jacobtylerwalls:
```suggestion
But this breaks when we are dealing with multi-column results. For example:
```

draft/table_value_expression.rst
```diff
@@ -0,0 +1,329 @@
...
+
+.. code-block:: python
+   first_object = Cohort.objects.order_by("id").values("subject", "duration")[:1]
+
+   Sales.objects.annotate(
+       first_subject=Subquery(first_object)
+   ).values("first_subject")
+
+This will raise `Cannot resolve expression type, unknown output_field`. bcs ORM don't know how to handle the multi-columns.
```
jacobtylerwalls:
```suggestion
This will raise `Cannot resolve expression type, unknown output_field`, because the ORM does not have an expression representing multiple columns.
```

draft/table_value_expression.rst
```diff
@@ -0,0 +1,329 @@
...
+   ).values("series")
+
+Conceptual SQL shape today:
+
+.. code-block:: sql
+
+   SELECT generate_series(1, 3) AS "series"
+   FROM "sales"
+
+it's not clear how to get it into the `FROM` clause.
```
jacobtylerwalls:
```suggestion
The ORM does not expose a way to place it in the `FROM` clause. 
```

draft/table_value_expression.rst
```diff
@@ -0,0 +1,329 @@
...
+The exact public API is still open.
+
+Expression-like table sources
+-----------------------------
+
+After output columns can be represented, the ORM needs a way to register an
+expression or query as a table source in ``FROM`` or ``JOIN``.
+
+We will take inspiration from existing `FilteredRelation`. We would follow the
+mechanism it used to register itself as a table source in ``FROM`` clause.
```
jacobtylerwalls: in -> in the

draft/table_value_expression.rst
```diff
@@ -0,0 +1,329 @@
...
+
+The exact public API is still open.
+
+Expression-like table sources
+-----------------------------
+
+After output columns can be represented, the ORM needs a way to register an
+expression or query as a table source in ``FROM`` or ``JOIN``.
+
+We will take inspiration from existing `FilteredRelation`. We would follow the
```
jacobtylerwalls: from -> from the

You might (?) also briefly mention why `FilteredRelation` is relevant -- it's the only way users have today to insert content into the JOIN clause (by appending a `WHERE` clause). Maybe it's fine.

draft/table_value_expression.rst
```diff
@@ -0,0 +1,329 @@
...
+------------------------------------------------
+
+The first question is the ORM capability Django needs:
+
+* represent table-like query sources,
+* know their output columns,
+* and resolve references to those columns.
+
+Once that capability is clear, the exact internal object model can be evaluated.
+
```
jacobtylerwalls:
```suggestion
```

jacobtylerwalls (approved):
Nice 👍 

I added minor wording suggestions.

@charettes, @LilyFirefly would you like to take a look?

draft/table_value_expression.rst
```diff
@@ -0,0 +1,323 @@
...
+The exact resolver integration point is open. Candidate locations include
+``Query.resolve_ref()``, ``Query.setup_joins()``, ``Query.names_to_path()``, or
+a smaller helper shared by multiple query paths.
+
+Rationale
+=========
+
+Why ``FilteredRelation`` is relevant
+------------------------------------
+It's the only way users have today to insert content into the ``JOIN`` clause (by appending a ``WHERE`` clause).
```
p-r-a-v-i-n: Should the position of this get change ?

draft/table_value_expression.rst
```diff
@@ -0,0 +1,323 @@
...
+But this breaks when we are dealing with multi-column results. For example:
+
+.. code-block:: python
+   first_object = Cohort.objects.order_by("id").values("subject", "duration")[:1]
+
+   Sales.objects.annotate(
+       first_subject=Subquery(first_object)
+   ).values("first_subject")
+
+This will raise `Cannot resolve expression type, unknown output_field`. because the ORM does not have an expression representing multiple columns.
```
charettes: Maybe add

```suggestion
This will raise `Cannot resolve expression type, unknown output_field`. because the ORM does not have an expression representing multiple columns and thus cannot make use of a join against the multi-column subquery.
```

draft/table_value_expression.rst
```diff
@@ -0,0 +1,323 @@
...
+Conceptual SQL shape today:
+
+.. code-block:: sql
+
+   SELECT ..., generate_series(1, 3) AS "series"
+   FROM "sales"
+   WHERE generate_series(1, 3) > 1
+
+The function is still not a table source. It is duplicated as an expression in
+the ``SELECT`` and ``WHERE`` clauses.
```
charettes: Should we explain why this is an issue with an inlined table of results and compare what the output would be if `series` was part of the `FROM` clause?

draft/table_value_expression.rst
```diff
@@ -0,0 +1,323 @@
...
+
+As a scalar expression, this has no single ``output_field``. As a table
+expression, it could expose named columns:
+
+.. code-block:: python
+
+   Sales.objects.alias(
+       cohort_data=TableExpression(
+           Cohort.objects.values("subject", "duration")
+       )
```
charettes: Not entirely sure we need a `TableExpression` wrapper in this case for same reason `Subquery` wrapping of `QuerySet` instances is unnecessary in most cases.

draft/table_value_expression.rst
charettes: Just to make sure I understand. You mean that we would only support usage for _annotation_ purpose in when building queries and not as a field that can be used when defining a model class?

draft/table_value_expression.rst
```diff
@@ -0,0 +1,323 @@
...
+-----------------------------
+
+After output columns can be represented, the ORM needs a way to register an
+expression or query as a table source in the ``FROM`` or ``JOIN``.
+
+We will take inspiration from the existing `FilteredRelation`. We would follow the
+mechanism it used to register itself as a table source in the ``FROM`` clause.
+
+This DEP uses ``TableExpression`` as a placeholder name. The final public API
+may use a different name.
```
charettes: I suspect we won't need `TableExpression` at all as we'll be able to perform the `FROM` aliasing based on other heuristics.

For example, in the case of `QuerySet.resolve_expression` if there are more than one field than it's `output_field` would be a `CompositeField` instance and we could make `GenerateSeries.output_field = CompositeField(IntegerField())` even if it's a singleton.

The heuristic to add a member to `sql.Query.alias_map` could then be `isinstance(output_field, CompositeField)`.

draft/table_value_expression.rst
```diff
@@ -0,0 +1,323 @@
...
+Example:
+
+.. code-block:: python
+
+   Sales.objects.alias(
+       series=TableExpression(GenerateSeries(1, 3))
+   ).filter(series__value__gt=1)
+
+For single-column table expressions, the final API may allow a shorter form such
+as ``series__gt=1``. That detail is still open.
```
charettes: That should be relatively straightforward to support by having `CompositeField.get_lookup` use it's only field when it's the case.

draft/table_value_expression.rst
charettes: I think it's worth pointing out, not sure where the best location is


```suggestion
Why ``FilteredRelation`` is relevant
------------------------------------
It's the only way users have today to insert content into the ``JOIN`` clause (by appending a condition to the ``ON`` clause).
```

draft/table_value_expression.rst
```diff
@@ -0,0 +1,323 @@
...
+   TableExpression(
+       JsonEach("payload"),
+       columns={
+           "key": models.TextField(),
+           "value": models.JSONField(),
+       },
+   )
+
+This may be easy to prototype, but it creates a separate metadata path from
+``Expression.output_field``.
```
charettes, line 510: Not convinced this is warranted or not something that could not be provided by `JSONEach` directly.

e.g.

```python
JsonEach("payload", key_column="attr")
```

could be turned into

```sql
(SELECT key AS "attr", value FROM json_each("payload"))
```

charettes (commented):
I left a few comments but overall the design document seems a like good foundation for further demonstration work of the viability of the approach.

I remain unconvinced of the necessity of the `TableExpression` abstraction as I believe `isinstance(expr, CompositeField)` (or a potential future `CompositeField` type narrowing `Field.is_composite -> bool` guard) is likely a good heuristic to fit our needs of augmenting `sql.Query.alias_map`. I believe this the case due to the complexity that `isinstance(expr, (Subquery, QuerySet, sql.Query))` caused when `Subquery` was introduced (e.g. `__in` lookups).

draft/table_value_expression.rst
p-r-a-v-i-n: We were also was in the favor of not adding `TableExpression` or something. But what we have discovered that `.alias_map` requires a `Table-like object`.  and it was failing on the line
 [            j for j in self.alias_map if self.alias_map[j].join_type == INNER](https://github.com/django/django/blob/f1440a752ec034277ccdad914995c3f164308e41/django/db/models/sql/query.py#L748)
 `Query` or `Subquery`  don't have `table_name`, `table_alias`, `join_type`, `parent_alias` which would require to live inside `alias_map` as per the [contract](https://github.com/django/django/blob/f1440a752ec034277ccdad914995c3f164308e41/django/db/models/sql/datastructures.py#L34-L43).
 
we might don't need `objects.alias(TableExpression(Subquery(...))` instead 

```
if isinstance(expr.output_field, CompositeField):
alias_map[alias] =TableExpression(expr)
```
also query object `as_sql` will generate something `(SELECT ... FROM ...)`  i'm assuming we might need something in prefix or suffix in the FROM clause with another table. for that also we need some handling.

draft/table_value_expression.rst
p-r-a-v-i-n: I have refined it in latest commits.

p-r-a-v-i-n:
> Just to make sure I understand. You mean that we would only support usage for annotation purpose in when building queries and not as a field that can be used when defining a model class?

yes

obstacles  I see without Table-Expression
for Subqueries:
1.  Query set don't have spacesuite to live inside the `.alias_map`. that spacesuite could be a Tableexpression.
2.  We might need to add some prefix or suffix to sql generated by subquery like  prefix `,` for cross join or something else

draft/table_value_expression.rst
charettes: I'm not sure I understand why the need for objects in `alias_map` to have certain attributes requires that objects passed to `QuerySet.alias` be `TableExpression`.

Why can't `alias` do something like

```python
if isinstance(expr.output_field, CompositeField):
   join = Join(
       expr,
       alias,
       ...
   )
   self.join(join)
```

In other words, shouldn't we adapt `Join` and `BaseTable` to be more _expression like_ and allow composition of other `Expression` over introducing `TableExpression` to bridge between them?

jacobtylerwalls: I overlooked that this version of the DEP still suggested that `TableExpression` was going to be exposed to users. As you mentioned in https://github.com/django/deps/pull/113#discussion_r3409860413 with the analogy to `Subquery`, it should be unnecessary in most instances, perhaps only if you need to specify a special kind of join?

We were just following the implementation hint in https://code.djangoproject.com/ticket/37115#comment:8 to create a base class for `BaseTable` and `Join`.

> requires that objects passed to QuerySet.alias be TableExpression.

I think @p-r-a-v-i-n is aware that we don't want users to have to supply these wrappers, see https://github.com/django/deps/pull/113#discussion_r3411094546, but we definitely need to update the DEP, because it suggests Simon's reading.

> In other words, shouldn't we adapt Join and BaseTable to be more expression like and allow composition of other Expression over introducing TableExpression to bridge between them?

I think this is where we're all heading, let's discuss once there's an implementation PR to look at, which I think is imminent.

Genarito:
Hi! Hope you're doing well!
Sorry if this is a silly question, but wouldn't this DEP conflict with the one I opened a few months ago?

https://github.com/django/deps/pull/106

jacobtylerwalls:
(Above, @p-r-a-v-i-n is using "spacesuit" as a colloquialism for "protocol".)

jacobtylerwalls:
> Just to make sure I understand. You mean that we would only support usage for annotation purpose in when building queries and not as a field that can be used when defining a model class?

I encouraged @p-r-a-v-i-n to defer that aspect for now, but we can re-prioritize it if you think it's important to include earlier. Can you say a little bit about the use cases for that?

charettes:
@jacobtylerwalls 

> I encouraged @p-r-a-v-i-n to defer that aspect for now, but we can re-prioritize it if you think it's important to include earlier.

I think it's smart to defer the implementation as it should cut scope significantly. I think that making sure `CompositePrimaryKey` subclasses `CompositeField` which subclasses `Field` would be beneficial though. We could make a `CompositeField.contribute_to_class` override that makes sure to raise an exception if `type(self) is CompositeField` to disallow this usage for now.

> Can you say a little bit about the use cases for that?

From my understanding that was the original intent behind #17279 / ticket-373 but we took a shortcut to focus solely on composite primary key at first (leaving us with the composite fk problem for example).

There was a SoC project for `CompositeField` [14 years ago](https://code.djangoproject.com/ticket/373#comment:103) and many discussions of the benefit of allowing _virtual_ field composition to exist.

draft/table_value_expression.rst
charettes: > it should be unnecessary in most instances, perhaps only if you need to specify a special kind of join?

right but I wonder if this should be a concern of this work or not, I think that laying down the foundation for table-like expressions is already a ton of work in itself.

> I think this is where we're all heading, let's discuss once there's an implementation PR to look at, which I think is imminent.

Sounds good! Peeking at code will definitely help here.

draft/table_value_expression.rst
```diff
@@ -0,0 +1,339 @@
...
+==========
+In this section we try to expose the problem we currently have
+For example, a scalar subquery can be used as an annotation:
+
+.. code-block:: python
+
+   first_subject = Cohort.objects.order_by("id").values("subject")[:1]
+
+   Sales.objects.annotate(
+       first_subject=Subquery(first_subject)
```
jacobtylerwalls: Let's remove the explicit `Subquery` throughout, as it may be distracting for readers who are used to eliding it.
```suggestion
       first_subject=first_subject
```

LilyFirefly:
> Hi! Hope you're doing well! Sorry if this is a silly question, but wouldn't this DEP conflict with the one I opened a few months ago?
> 
> #106

There's definitely some overlap, but this DEP is more about laying the groundwork for a wide collection of features, including CTEs. If this GSOC work goes well, you may be able to rewrite #106 to build on top of this work.

draft/table_value_expression.rst
```diff
@@ -0,0 +1,339 @@
...
+   Cohort.objects.values("subject", "duration")
+
+As a scalar expression, this has no single ``output_field``. As a table
+expression, it could expose named columns:
+
+.. code-block:: python
+
+   Sales.objects.alias(
+       cohort_data=TableExpression(
+           Cohort.objects.values("subject", "duration")
```
LilyFirefly: Is the `[:1]` deliberately omitted in the second example? My understanding is that the `[:1]` makes sure there's only one row selected from the subquery, but the examples here are focused on the case of more than one column, so the `[:1]` would be appropriate in both cases.

p-r-a-v-i-n: that's fair. thanks

draft/table_value_expression.rst
jacobtylerwalls: The question about join type came up right away in the first cut of the implementation, any guidance here @charettes and @LilyFirefly would be very helpful!

https://github.com/django/django/pull/21496#discussion_r3455381877

draft/table_value_expression.rst
```diff
@@ -0,0 +1,336 @@
...
+
+The actual SQL generated:
+
+.. code-block:: sql
+
+   SELECT "SRF_sales"."id", "SRF_sales"."sold_on", "SRF_sales"."day", generate_series(1, 3) AS "series"
+   FROM "SRF_sales"
+
+
+The ORM does not expose a way to place the set-returning function in the ``FROM`` clause.
```
LilyFirefly, line 97: It would be nice to also include the SQL we want to be able to generate, so readers can compare.

p-r-a-v-i-n: Thanks. I have added it now.

draft/table_value_expression.rst
```diff
@@ -0,0 +1,342 @@
...
+In this example, the ORM cannot reliably know if the ``subquery`` returns exactly one row or multiple rows.
+To solve this, our approach requires the developer to explicitly declare when a result is a multi-row table source.
+The user will do this by wrapping the queryset in a ``TableExpression``.
+
+.. code-block:: python
+
+   User.objects.alias(posts=TableExpression(subquery))
+
+This explicit wrapper allows the ORM to cleanly separate the processing logic. When encountering a ``TableExpression``, the ORM knows to place the subquery inside the ``FROM`` clause (utilizing ``LATERAL`` joins if there are outer references) rather than resolving it as a scalar subquery in the ``SELECT`` clause.
+*(Note: This DEP uses ``TableExpression`` as a placeholder name. The final public API may use a different name).*
```
p-r-a-v-i-n: Please don't mind for writing very late and the reason is i'm being shy to write , I'm afraid of if I sound like "I'm forcing something or being very concrete on something".  

I have try to add clarity in this section why do we might need `TableExpression` wrapper.

 @jacobtylerwalls @LilyFirefly @charettes

draft/table_value_expression.rst
p-r-a-v-i-n: yes we don't in this case. Thanks.  But the necessity i try to explain in the  section mention in this [comment](https://github.com/django/deps/pull/113/changes#r3537680606)

draft/table_value_expression.rst
LilyFirefly: Wouldn't the ~scalar~ *single row* case look like: (note the `[:1]`)

```python
   subquery = Post.objects.filter(user=first_user).values("title", "body")[:1]
   User.objects.alias(posts=subquery).values("posts__title", "posts__body")
```

Or am I missing something here? (I might be!)

p-r-a-v-i-n: Please correct me, I'm assuming you doubt that i have used similar example with `[:1]` in scaler case.

below is just examples  (my understanding) , so we both are on same page

> subquery = Post.objects.filter(user=first_user).values("title")[:1]

This one is scaler.

> subquery = Post.objects.filter(user=first_user).values("title", "body")[:1]

This is multi-column single row

> subquery = Post.objects.filter(user=first_user).values("title", "body")

If `first_user` has created multiple posts the this  will return  multi-column multi-row

i used this so it represent the `table-value` result

jacobtylerwalls: Ah, so @LilyFirefly is your idea that we can inspect whether the queryset is sliced, and if not, treat it as multi-row, without the need for a wrapper?

LilyFirefly: I think it's consistent with how we've designed subqueries up to this point and I think it's worth exploring if it will work in practice or cause problems (maybe it's too subtle and people will get confused?).

jacobtylerwalls: I like it -- let's explore it.

draft/table_value_expression.rst
```diff
@@ -0,0 +1,536 @@
...
+``Query.alias_map``. It reuses that join or creates one when needed.
+
+The derived table is then added to the ``FROM`` clause:
+
+.. code-block:: sql
+
+   LEFT OUTER JOIN (
+       SELECT email, name
+       FROM ...
+   ) info ON (1 = 1)
```
LilyFirefly, line 389: Do we *need* the `ON (1 = 1)`? Might future work want more complex `ON` clauses?

draft/table_value_expression.rst
```diff
@@ -0,0 +1,536 @@
...
+
+Join type follows normal queryset boolean semantics. When a table source is
+required by the query, it is compiled as a ``CROSS`` join. If the function returns
+no rows, the corresponding base row also produces no result:
+
+.. code-block:: python
+
+   Sales.objects.alias(
+       series=GenerateSeries(1, 3),
+   ).filter(series__gt=1)
```
LilyFirefly: I think this example would benefit from showing which `Sales` objects and `series` values are returned and making it clear which are filtered out. Currently the reader has to guess at how this snippet behaves.

draft/table_value_expression.rst
p-r-a-v-i-n: Yes, we need it because the subquery is an independent table source and there is no condition connecting it to the outer table. `ON (1 = 1)` makes every inner row available to every outer row, while` LEFT OUTER JOIN` preserves outer rows when the subquery is empty. In future, if we support a real join condition, we could extend it to accept it .

draft/table_value_expression.rst
jacobtylerwalls: @p-r-a-v-i-n 👀

PR #114, opened 2026-06-18
Experimental APIs

prafulgulani:
This DEP proposes "Experimental" space to document, test and mature unstable features before they are locked into permanent stability guarantee.

References:
- Proposed as part of GSoC 2026
- Discussion thread in new-features: https://github.com/django/new-features/issues/3
- Forum: https://forum.djangoproject.com/t/implementing-a-formal-experimental-api-for-django/45217

### DIFF
--- a/draft/experimental-apis.rst
+++ b/draft/experimental-apis.rst
@@ -1,0 +1,200 @@
+==============================
+DEP XXXX: Experimental APIs
+==============================
+
+:DEP: XXXX
+:Author: Andrew Miller, Praful Gulani
+:Implementation Team: Andrew Miller, Praful Gulani
+:Shepherd: Andrew Miller
+:Status: Draft
+:Type: Process
+:Created: 2026-06-08
+:Replaces: `DEP 2 <https://github.com/django/deps/blob/main/draft/0002-experimental-apis.rst>`_.
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+Abstract
+========
+
+This DEP proposes a formal process for introducing "Experimental" features and 
+APIs into Django. Currently, Django operates on an all-or-nothing model: features 
+are either hidden and undocumented, or fully public and locked into a strict, 
+multi-year stability guarantee. While this approach ensures that official releases 
+are exceptionally stable, it struggles when features carry a unique set of challenges.
+Challenges like API design deadlocks, partial implementations that need to land 
+incrementally, or updates to core systems like auth.User that lead to breaking 
+changes, require an alternate approach.
+
+This proposal creates an isolated space for the features to evolve under 
+real-world usage and feedback, all while keeping production environments safe from 
+accidental instability. Features in this space will be fully documented so they 
+are visible to the community, but they will remain explicitly exempt from standard
+backwards-compatibility policy until they mature.
+
+Motivation
+==========
+
+Django values stability and strict backwards compatibility the most.
+Under the current policy, the moment a new feature is documented and released, it is 
+locked into a multi-year deprecation cycle. Because of this, the community needs 
+absolute design perfection before code can be merged. While this approach works 
+great and is the reason behind Django's stability, it has also led to prolonged 
+debates for some features which has caused some valuable contributions to lose momentum.
+
+The current methods to bring new features into Django work fine most of the time.
+However, when a contribution introduces a complex change, it can be difficult to 
+make progress using current approaches. The new experimental path serves as an 
+alternate option specifically for scenarios where the existing methods face limitations:
+
+1. Third-Party Packages
+-----------------------
+Third-party packages work great for adding new features or tools. However, when a feature
+fundamentally changes Django's internal mechanics or early startup steps, an alternate
+approach is needed to test those changes safely.
+
+We see this in `Ticket-24312 (making it possible to import models safely at any time) <https://code.djangoproject.com/ticket/24312>`_.
+Rewriting Django’s app-loading system to remove its reliance on import side 
+effects is a massive architectural shift. A third-party package cannot solve 
+this because it is bound by the very loading system it tries to fix. An experimental 
+namespace provides a safe playground to map out what the final design of such a 
+core refactor looks like without breaking stable production environments.
+
+2. Merging Directly into Core
+-----------------------------
+Django requires wide compatibility across all supported backends and 
+environments before code is merged into the main codebase. However, demanding 
+simultaneous perfection across every single environment on day one can sometimes
+delay mature parts of a feature from being integrated. Furthermore, attempting 
+to land massive changes all at once places a very heavy burden on both code 
+submitters and reviewers.
+
+We see this in `Ticket-28805 (adding database functions for regular expressions) <https://code.djangoproject.com/ticket/28805>`_. 
+While implementations for some database backends were ready, the feature stalled for 
+years because core integration required simultaneous support across every single 
+environment. An experimental space would allow ready components to land, letting the 
+community use those features while the remaining database compatibility issues are
+resolved over subsequent releases.
+
+3. Undocumented Features
+------------------------
+Maintainers sometimes choose not to document temporary extension points or core 
+refactors because the design is not yet stable enough for a permanent public 
+commitment. While this successfully protects production environments from breaking
+changes, it naturally limits community awareness. Without documentation, only a 
+small group of developers ever discover them and provide feedback.
+
+An Experimental Space
+---------------------
+
+This DEP introduces an "Experimental" space to create a safe, native pathway to 
+improve framework features. By letting code land without immediate, permanent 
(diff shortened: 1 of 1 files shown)


draft/experimental-apis.rst
```diff
@@ -0,0 +1,200 @@
...
+stability promises, maintainers get the flexibility to iterate on complex designs 
+based on real-world usage and feedback.
+
+This isolated space will also allow us to safely evolve existing 
+core systems. It provides a safe zone to experiment with modernizing critical, 
+long-standing framework components without breaking existing projects. For example, 
+`updating auth.User <https://buttondown.com/carlton/archive/evolving-djangos-authuser/>`_ 
+to replace the ``first_name`` and ``last_name`` fields with a singular ``name`` field, 
+or making the email unique, becomes possible because the community can natively test 
+the changes before they are made permanent.
```
LilyFirefly, line 102: While I'd love to see the `first_name` and `last_name` go, I'm not clear enough on how the experimental zone will interact with migrations to enable this cleanly.

Would we have new migrations in the experimental area that we opt-in to by adding a suitable entry in `INSTALLED_APPS`? If so, what would the process be for moving the migrations into stable once the experiment is successful? How do we handle removing migrations if the experiment is unsuccessful?

draft/experimental-apis.rst
```diff
@@ -0,0 +1,200 @@
...
+
+1. Community Support Check:
+   Every feature begins with a proposal in the django/new-features repository 
+   to gather initial community feedback and see if the idea belongs in Django core.
+
+2. Selecting the Experimental Path:
+   During the "Can we do it?" phase, if the Steering Council or the author decides 
+   that a feature is too complex to merge as a stable release immediately either 
+   due to high compatibility risks or the need for multi-database testing, they can 
+   choose the experimental pathway.
```
LilyFirefly: I think the Steering Council would have final say on whether the experimental pathway is suitable. I'm not sure that saying "the author decides" is quite right.

LilyFirefly (commented):
Thanks for making a start on this! I have a bit of early feedback:

draft/experimental-apis.rst
nanorepublica: Very good questions Lily! I suggested this area as an example for us to ground the conversation of the flags in a real problem which the community would like to see changed, but very likely impossible to change via the current methods (I could be wrong). 

While the migrations would be tricky, it enables to propose a fresh default User model to work from (at least for new projects).

More broadly this is something we (myself & Praful) can experiment with once we something of flags up and running.

draft/experimental-apis.rst
```diff
@@ -0,0 +1,200 @@
...
+Django requires wide compatibility across all supported backends and 
+environments before code is merged into the main codebase. However, demanding 
+simultaneous perfection across every single environment on day one can sometimes
+delay mature parts of a feature from being integrated. Furthermore, attempting 
+to land massive changes all at once places a very heavy burden on both code 
+submitters and reviewers.
+
+We see this in `Ticket-28805 (adding database functions for regular expressions) <https://code.djangoproject.com/ticket/28805>`_. 
+While implementations for some database backends were ready, the feature stalled for 
+years because core integration required simultaneous support across every single
```
jacobtylerwalls, line 75: I don't draw this same conclusion when I read the ticket. I think it stalled because of the complexity of "juggling so many database-specific flags", which is the nature of a cross-database abstraction given the complexity of the function we are talking about. I don't think this example backs up the "why" point here.

draft/experimental-apis.rst
```diff
@@ -0,0 +1,200 @@
...
+to replace the ``first_name`` and ``last_name`` fields with a singular ``name`` field, 
+or making the email unique, becomes possible because the community can natively test 
+the changes before they are made permanent.
+
+By having this isolation space, we wouldn’t have to change Django’s policy regarding 
+undocumented features. Instead, by bringing unstable components under the experimental
+space, we can safely document them and provide a direct link to a dedicated forum
+discussion where all the feedback can be gathered.
+
+The Experimental Feature Pipeline
```
nanorepublica, line 109: One broad topic I come back to is with this idea is how this works with the current release cadences and how the community installs Django.

Currently Django has 3 release types. Bugfix, Feature & Series releases (from DEP 19). DEP 20 changes this and effective drops this down to 2 types (Feature & Bugfix). 

Bugfix releases are shipped monthly. This is the current minimum time to ship something as I understand it, but these bugfixes are specifically backported to frozen feature branches. 

I think therefore there is only one way for developers to test these experimental features quickly inside Django and that is installing Django from the main branch (or a slightly delayed main branch). - which is not a bad thing, ideally more folks would test/use Django against main (https://noumenal.es/posts/running-django-main/O2x/)

This means only those installing from main will get experimental features iteratively (and therefore quickly)

---

Worth noting the above is my current thoughts. I'm working through how this affects the project overall and being useful to the community

nanorepublica: The other possible path here is that we choose to start using python package extras to generate releases with the experimental code included. This would mean that we wouldn't be limited to having code within Django core that appears in a final Django package. 

I generally think that we need to investigate the release process and what the Fellows would think to additional work added to the release process or more additional releases being added to the process.

draft/experimental-apis.rst
```diff
@@ -0,0 +1,200 @@
...
+:Replaces: `DEP 2 <https://github.com/django/deps/blob/main/draft/0002-experimental-apis.rst>`_.
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+Abstract
+========
+
+This DEP proposes a formal process for introducing "Experimental" features and
```
tim-schilling, line 21: Low: I think the term "experimental" needs clarification. I suspect folks have different opinions on it. For me, I imagine it's something that's fairly likely to be merged in, but there's _some_ hesitancy. It's basically just a new stage of integrating a feature, but in this case we know we need to go slower with this feature than normal.

I'm drawn to multi-column primary keys as something that may have benefited from this. It's implemented, but not fully. Would it be helpful to be able to change the public API for another release without deprecations?

draft/experimental-apis.rst
```diff
@@ -0,0 +1,200 @@
...
+   included in that release.
+
+*  **Documenting What is Missing:** Documentation for experimental components must not 
+   just state how the feature works, it should also explicitly list what is currently 
+   missing or incomplete, so users know exactly where to focus their testing efforts.
+
+*  **Direct Links for Forum Feedback:** Every experimental documentation block will 
+   include a direct link pointing to a dedicated feedback thread on the Django Forum. 
+   This ensures all user experiences, bug reports, and design suggestions are collected 
+   in one place.
```
tim-schilling, line 155: I think this needs rethinking. Bug reports should always go through Trac. Being able to note the experiment or feature in there would be helpful.

I think the feedback portion is a good idea, but we should provide a bit more structure for someone. If we have an experiment, there was a reason for it. That's the spot that we particularly want feedback on. How do we make that easier for everyone?

draft/experimental-apis.rst
tim-schilling: Would multi-column primary keys be a more relevant example for this?

draft/experimental-apis.rst
tim-schilling: > I think therefore there is only one way for developers to test these experimental features quickly inside Django and that is installing Django from the main branch (or a slightly delayed main branch). - which is not a bad thing, ideally more folks would test/use Django against main

Recommending people to use main for production applications is probably not a good idea. Mistakes still happen. For a project that hangs its hat on stability, recommending people use main in production seems like contradiction. Sure, there's room to have "versions" of main that are more reliable or have a high degree of reliability, but that doesn't exist in our process today so it likely shouldn't be included here.

draft/experimental-apis.rst
jacobtylerwalls: Yes!

jacobtylerwalls: Well, it has a different effect than the original example because it's something that we *did* merge without putting it under an experimental namespace and without the need for shipping a lot of incremental deprecations for design changes. So its success doesn't really demonstrate the need for a change.

tim-schilling: Sure, but I meant it more as could have things gone better with a less restrictive support policy for it? Better being subjective across a number of facets (communication, documentation clarity, development, etc)

LilyFirefly:
I wonder if support for http `QUERY` (https://github.com/django/new-features/issues/185) would be a good candidate for exposing via experimental APIs?

nanorepublica:
Ha 😄 , I was thinking the same thing given that precise feature has come up 4 separate times now. 

It's something that would be appropriate for core and would allow us to develop and release it while the RFC is being still proposed. Only once it becomes finalised do we make it stable.

PR #115, opened 2026-08-05
Use X in Deprecation Policy section of DEP 20 for clarity.

LilyFirefly:


carltongibson (approved):
Yes, good. Thanks @LilyFirefly — I only caught it in the Rationale section 👍

PR #118, opened 2026-08-21
Adding Third-Party Packages as Package Extras

prafulgulani:
This DEP proposes adding third-party packages as package extras.

References:
- Proposed as part of GSoC 2026
- Discussion thread in new-features: https://github.com/django/new-features/issues/3

### DIFF
--- a/draft/package-extras.rst
+++ b/draft/package-extras.rst
@@ -1,0 +1,104 @@
+=======================================================
+DEP XXXX: Adding Third-Party Packages as Package Extras
+=======================================================
+
+:DEP: XXXX
+:Author: Praful Gulani, Andrew Miller
+:Implementation Team: Praful Gulani, Andrew Miller
+:Shepherd: Andrew Miller
+:Status: Draft
+:Type: Process
+:Created: 2026-08-21
+
+.. contents:: Table of Contents
+   :depth: 3
+   :local:
+
+Abstract
+========
+
+While Django follows a batteries-included philosophy, including every new 
+feature/requirement according to modern application demands would lead to 
+bloated core and increased load on maintainers. Using package extras would 
+keep Django core lean and would also increase the visibility of high quality 
+third-party packages. 
+
+This DEP defines the requirements and testing needed for adding third-party 
+packages as "package extras" within the django framework. The installation of 
+package extras shouldn’t affect django’s stability and performance and packages 
+need to guarantee this by maintaining appropriate isolation and bridging test 
+suites for flawless integration. 
+
+
+Motivation
+==========
+
+Django’s batteries included approach makes it a reliable framework for web 
+development. But the web ecosystem is huge and as it evolves it can be tough to 
+incorporate everything directly into Django core. Trying to give everything out 
+of the box can lead to bloated codebase and can increase the burden on maintainers.
+
+Django is also known for its massive and robust third party package ecosystem, 
+most django projects use some third-party package for essential functionality. 
+A lot of well maintained third-party packages receive the attention they deserve 
+but many of them don’t. 
+
+By introducing package extras the whole community gets benefited:
+* By not adding everything to the core we can keep the core lean and also allow 
+the functionality to be added officially without any maintenance overhead.
+* Listing well maintained third-party packages as extras increases their 
+visibility for the users and also acts as encouragement for the developers.
+
+
+Guidelines for third party packages
+===================================
+
+* The package must have a documented security policy and a responsive process 
+  for reporting vulnerabilities.
+* The package must use a license compatible with Django's BSD-3 clause license.
+* Top-level ``__init__.py`` files must not perform eager network calls, heavy 
+  IO, or database access.
+* The package must follow a deprecation schedule aligned with Django's releases.
+* Packages must use lower-bound constraints (``package>=X.Y``) rather than exact 
+  versions (``package==X.Y.Z``) for declaring Django dependency.
+* Third-party packages that rely on private Django APIs must run CI matrix 
+  checks against ``django/main``.
+
+
+Specification
+=============
+
+Listing package extras without a proper process introduces significant risks:
+* Some third-party packages might alter global settings with a simple install 
+before being configured in ``settings.py``
+* Poorly specified or circular third-party dependencies can break installations 
+across package managers.
+* Unconfigured extras executing early imports might lead to performance degradations.
+
+To safely list package extras without compromising the stability and performance 
+of Django we need to have a test suite that helps with isolation and bridging 
+of the packages.
+
+
+Rationale
+=========
+
+This section should flesh out the specification by describing what motivated 
+the specific design and why particular design decisions were made. It should 
+describe alternate designs that were considered and related work.
+The rationale should provide evidence of consensus within the community and 
+discuss important objections or concerns raised during discussion.
+
+
+Backwards Compatibility
+=======================
+
+This DEP doesn't introduce any backwards incompatibilities. It does not modify 
+any of the existing django applications.
+
+
+Copyright
+=========
+
+This document has been placed in the public domain per the Creative Commons 
+CC0 1.0 Universal license (https://creativecommons.org/publicdomain/zero/1.0/deed).


jonbiemond:
Great work! Does "Package Extras" refer to [optional dependencies](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#extras) or something more conceptual?

prafulgulani:
> Great work! Does "Package Extras" refer to [optional dependencies](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#extras) or something more conceptual?

Thanks! Yes, it refers to optional dependencies in pyproject.toml so people can run pip install django[abc].
But in a little different context, this isn't about listing jinja2, numpy, Pillow etc., that has been discussed [here](https://github.com/django/new-features/issues/157) and also has a PR.
This is about listing something like experimental as optional dependencies, you might want to take a look at [experimental-apis DEP](https://github.com/django/deps/pull/114).

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+but many of them don’t. 
+
+By introducing package extras the whole community gets benefited:
+* By not adding everything to the core we can keep the core lean and also allow 
+the functionality to be added officially without any maintenance overhead.
+* Listing well maintained third-party packages as extras increases their 
+visibility for the users and also acts as encouragement for the developers.
+
+
+Guidelines for third party packages
```
tim-schilling, line 53: This is likely better off in the specification section.

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+Listing package extras without a proper process introduces significant risks:
+* Some third-party packages might alter global settings with a simple install 
+before being configured in ``settings.py``
+* Poorly specified or circular third-party dependencies can break installations 
+across package managers.
+* Unconfigured extras executing early imports might lead to performance degradations.
+
+To safely list package extras without compromising the stability and performance 
+of Django we need to have a test suite that helps with isolation and bridging 
+of the packages.
```
tim-schilling, line 80: Should we be running smoke tests with the extras in the Django CI to confirm when a problem may crop up based on a change to core?

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Specification
+=============
+
+Listing package extras without a proper process introduces significant risks:
+* Some third-party packages might alter global settings with a simple install 
+before being configured in ``settings.py``
+* Poorly specified or circular third-party dependencies can break installations 
+across package managers.
+* Unconfigured extras executing early imports might lead to performance degradations.
```
tim-schilling, line 76: I think a fair amount in the current specification section probably belongs in rationale. Or maybe I'm misunderstanding the purpose here?

Specifically, these two sentences make partial claims without evidence or examples which makes it hard for others to know what they should avoid explicitly. I feel like the specification section should be pretty clear on how extras are decided, implemented and maintained.

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Guidelines for third party packages
+===================================
+
+* The package must have a documented security policy and a responsive process 
+  for reporting vulnerabilities.
+* The package must use a license compatible with Django's BSD-3 clause license.
+* Top-level ``__init__.py`` files must not perform eager network calls, heavy 
+  IO, or database access.
+* The package must follow a deprecation schedule aligned with Django's releases.
```
tim-schilling, line 61: I'd like to see more about what is an aligned release schedule.

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+* Packages must use lower-bound constraints (``package>=X.Y``) rather than exact 
+  versions (``package==X.Y.Z``) for declaring Django dependency.
+* Third-party packages that rely on private Django APIs must run CI matrix 
+  checks against ``django/main``.
```
tim-schilling, line 65: Probably worth us requiring that the tests should run weekly as well as on PR/push, just in case their development cycle is slower than Django's. Especially if we don't have a smoke test in core for the packages.

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+the functionality to be added officially without any maintenance overhead.
+* Listing well maintained third-party packages as extras increases their 
+visibility for the users and also acts as encouragement for the developers.
+
+
+Guidelines for third party packages
+===================================
+
+* The package must have a documented security policy and a responsive process 
+  for reporting vulnerabilities.
```
tim-schilling, line 57: Do we need anything about governance for the projects? A project like Mongo has a company behind it which is much different scenario than a project that has a single developer with the only commit bit.

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+the specific design and why particular design decisions were made. It should 
+describe alternate designs that were considered and related work.
+The rationale should provide evidence of consensus within the community and 
+discuss important objections or concerns raised during discussion.
+
+
+Backwards Compatibility
+=======================
+
+This DEP doesn't introduce any backwards incompatibilities. It does not modify
```
tim-schilling, line 96: Do we need to note that these extras are only available for certain versions of Django?

draft/package-extras.rst
```diff
@@ -0,0 +1,104 @@
...
+* Some third-party packages might alter global settings with a simple install 
+before being configured in ``settings.py``
+* Poorly specified or circular third-party dependencies can break installations 
+across package managers.
+* Unconfigured extras executing early imports might lead to performance degradations.
+
+To safely list package extras without compromising the stability and performance 
+of Django we need to have a test suite that helps with isolation and bridging 
+of the packages.
+
```
tim-schilling, line 81: Should we include any information on why an extra may be removed?

tim-schilling (commented):
Thank you for the work on this @prafulgulani! I've got some questions around things we may want to include and suggestions on where things should live in the document. I think you're headed in the right direction defining the limits and requirements for packages. That will communicate things clearly which is always helpful 😊

draft/package-extras.rst
prafulgulani: We definitely need to define governance criteria so that we can avoid removing package as much as possible.
Few requirements we can consider could be:
1. We can mandate at least 2 maintainers.
2. We can check if they support every stable version of Django.
3. We can look at release track record to check for timely release for new versions of Django.
4. We need to look at how useful a particular package is in order to avoid adding a lot of packages.
what else do you think we should look for in a package before we add them?

draft/package-extras.rst
prafulgulani: It's not necessary for a package to move at the same pace as Django. They might release more or fewer versions than Django does, but irrespective of their release cycle, they should maintain compatibility with older Django releases for as long as Django supports those versions. When Django deprecates an API the package depends on, the package shouldn't instantly drop support for it.

And they should also test against Django's alpha/beta/RC releases and ship a compatible release within a reasonable timeframe, like 30 days or so.

draft/package-extras.rst
prafulgulani: Yes, it looks mixed as of now, what I had in mind was this section along with examples would serve as what should be avoided and I ended up writing these in specification. 
I'll restructure both the sections accordingly.

draft/package-extras.rst
prafulgulani: Yes, as much as we want to avoid removing them, we might need to remove a package someday, so a section outlining possible reasons would be an excellent addition.

draft/package-extras.rst
prafulgulani: I didn't get this! Do you mean that extras will be available from the versions that come after this DEP is accepted? or something else?

tim-schilling: If we have `django[db-vendor]` has been working for up to 2030 and then we have to remove it, is there anything we should consider up front now? For example, do we have a special deprecation process for removing an extra? Now that I'm writing it out, it seems like this concern should be handled elsewhere and not in the Backwards Compatibility section.

draft/package-extras.rst
tim-schilling: > It's not necessary for a package to move at the same pace as Django. 
...
> And they should also test against Django's alpha/beta/RC releases and ship a compatible release within a reasonable timeframe, like 30 days or so.

Yeah, this is the discussion I was hoping to extract out. I feel like a month is fast for OSS, but we could always loosen this too if it is problematic.

draft/package-extras.rst
tim-schilling: I think there are individuals who have been maintaining packages for 5+ years that I would be fine adding as an extra. Not that the package is 5 years old, but that they've been a contributor for that long. I think we could carve out exceptions where needed. Otherwise, agreed on the multiple maintainers bit. Or some other proof of governance that shows it's unlikely to disappear.

Agreed on 2 and 3 entirely.

Regarding 4, we can use is @adamchainz [well-maintained test](https://adamj.eu/tech/2021/11/04/the-well-maintained-test/).

[I've also been working on a project](https://github.com/tim-schilling/django-probe) to allow us to gather this type of metric based on several conversations. Not that it solves it now, but we could use this as a objective way to see what third-party packages are actually in use by the community.

tim-schilling:
Hi @prafulgulani and @nanorepublica, what about reducing scope on this to just those packages that we already define in the docs, such as [psycopg](https://docs.djangoproject.com/en/6.1/ref/databases/#postgresql-notes) and allowing for the option for the board to have a corporate sponsorship grant an extra?

This isn't coming from the SC, but from me as a community member. I was in the DSF office hours meeting this week and it was brought up that having the ability to have an extra may be appealing to a theoretical large corporate sponsor. Reducing the scope would allow for us to get some extras out into the wild sooner and learn from that experience rather than trying to get it right for all cases immediately.

I think this would table [my concern about governance](https://github.com/django/deps/pull/118#discussion_r3890199690) to a later date. It'd potentially remove a lot of the concerns and text around third-party packages being extensions of Django, simplifying the DEP.

jacobtylerwalls:
SGTM, and in the smaller-scope case we should be able to skip writing up a DEP, as we already have [ticket-28905.](https://code.djangoproject.com/ticket/28905)

PR #121, opened 2026-09-18
Remove the django-admin runtime notice

ryanhiebert:
Based on [feedback from the implementation PR](https://github.com/django/django/pull/21260#issuecomment-5700642421) for the new django command, remove the requirement for django-admin to print a notice about the new command name. The alias will remain indefinitely and the notice interferes with shell completion, so it doesn't seem worth the cost.


tim-schilling (approved):
Thanks for circling back on this. Do you expect additional changes or should we plan on merging this when others get eyes on it?

ryanhiebert:
I don't currently expect other changes, but if we want to hold on merging until the PR is finalized, that's fine with me.

adamchainz (approved):
Seems fair.

